PR #87, opened 2022-09-16
Wrong hash and no release errors 

alvarocabanas:
- Error if hash commit doesn't exist.
- If there are no tags, calculate next version from 0.0.1

### DIFF
--- a/src/app/nextversion/nextversion_test.go
+++ b/src/app/nextversion/nextversion_test.go
@@ -257,19 +257,6 @@ changes:
 			`),
 		},
 		{
-			name:     "Next_When_No_Tags_Only_Deps",
-			expected: "v0.1.0",
-			tags:     nil,
-			yaml: strings.TrimSpace(`
-notes: ""
-changes: []
-dependencies:
-- name: foobar
-  from: 0.0.1
-  to: 0.1.0
-			`),
-		},
-		{
 			name:     "Major_Capped_To_Minor",
 			expected: "v2.1.0",
 			args:     "--bump-cap=minor",
@@ -329,6 +316,30 @@ dependencies:
 - name: foobar
   from: 0.0.1
   to: 1.0.0
+			`),
+		},
+		{
+			name:     "Next_When_No_Tags",
+			expected: "v0.0.1",
+			tags:     nil,
+			yaml: strings.TrimSpace(`
+notes: ""
+changes:
+- type: breaking
+  message: I am marked as breaking but really I am first commit
+			`),
+		},
+		{
+			name:     "Next_When_No_Tags_Only_Deps",
+			expected: "v0.0.1",
+			tags:     nil,
+			yaml: strings.TrimSpace(`
+notes: ""
+changes: []
+dependencies:
+- name: foobar
+  from: 0.0.1
+  to: 0.1.0
 			`),
 		},
 	} {
--- a/src/app/nextversion/nextversion_test.go
+++ b/src/app/nextversion/nextversion_test.go
@@ -118,7 +118,7 @@ changes:
 
 //nolint:funlen,paralleltest // urfave/cli cannot be tested concurrently.
 func TestNextVersion(t *testing.T) {
-	tags := []string{
+	allTags := []string{
 		"v0.1.0",
 		"v1.0.0",
 		"v2.0.0", // Unordered on purpose, this should be the current version.
@@ -133,12 +133,14 @@ func TestNextVersion(t *testing.T) {
 		name     string
 		yaml     string
 		args     string
+		tags     []string
 		expected string
 	}{
 		{
 			name:     "Overrides_Next",
 			args:     "--next v0.0.1",
 			expected: "v0.0.1",
+			tags:     allTags,
 			yaml: strings.TrimSpace(`
 notes: |-
     ### Important announcement (note)
@@ -159,8 +161,8 @@ dependencies:
 		},
 		{
 			name:     "Overrides_Current",
-			args:     "--current v0.0.1",
-			expected: "v1.0.0",
+			expected: "v3.0.0",
+			tags:     allTags,
 			yaml: strings.TrimSpace(`
 notes: |-
     ### Important announcement (note)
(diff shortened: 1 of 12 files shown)


src/changelog/sources/dependabot/source_test.go
```diff
@@ -31,13 +36,16 @@ func (cl commitList) Commits(_ string) ([]git.Commit, error) {
 	return cl, nil
 }
 
-//nolint:funlen
+//nolint:funlen,maintidx
```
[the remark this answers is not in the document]
alvarocabanas: Fixed

src/changelog/sources/dependabot/source_test.go
```diff
@@ -202,6 +210,25 @@ Bumps [actions/github-script](https://github.com/actions/github-script) from 4.0
...
+			expected: []changelog.Dependency{{
+				Name: "github.com/spf13/viper",
+				From: semver.MustParse("1.7.0"),
+				To:   semver.MustParse("1.10.1"),
+				Meta: changelog.EntryMeta{
+					PR:     "",
+					Commit: "abcda222",
+				},
+			}},
+		},
```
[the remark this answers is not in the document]
alvarocabanas: Fixed

src/git/commit.go
```diff
@@ -50,25 +52,29 @@ func (s *RepoCommitsGetter) Commits(lastHash string) ([]Commit, error) {
...
 		cm, errCommit := commitIter.Next()
 		if errCommit != nil && !errors.Is(errCommit, io.EOF) {
 			return nil, fmt.Errorf("iterating git commits: %w", errCommit)
 		}
 
-		if prevWasLastHash || cm == nil {
+		// No more commits to iterate
+		if cm == nil {
+			if lastHash != "" {
+				return nil, fmt.Errorf("finding commits: %w", ErrNonexistentCommitHash)
```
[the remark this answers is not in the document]
alvarocabanas: Removed

src/changelog/sources/renovate/source.go
```diff
@@ -28,7 +29,7 @@ func NewSource(tagsVersionGetter git.TagsVersionGetter, commitsGetter git.Commit
 
 func (r Source) Changelog() (*changelog.Changelog, error) {
 	lastHash, err := r.tagsVersionGetter.LastVersionHash()
-	if err != nil {
+	if err != nil && !errors.Is(err, git.ErrNoReleases) {
```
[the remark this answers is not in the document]
alvarocabanas: Removed

src/bumper/bumper.go
```diff
@@ -65,3 +62,7 @@ func (b Bumper) BumpSource(source version.Source) (*semver.Version, error) {
 
 	return b.Bump(versions[0])
 }
+
+func (b Bumper) InitialVersion() *semver.Version {
+	return semver.MustParse("v0.0.1")
+}
```
roobre, line 68: Is there a reason for this to be a method and to be exported?

src/bumper/bumper.go
```diff
@@ -55,7 +52,7 @@ func (b Bumper) BumpSource(source version.Source) (*semver.Version, error) {
 	}
 
 	if len(versions) == 0 {
-		return nil, ErrNoTags
+		versions = []*semver.Version{b.InitialVersion()}
```
roobre, line 55:
```suggestion
		return semver.MustParse("v0.0.1"), nil
```

src/app/nextversion/nextversion_test.go
```diff
@@ -238,18 +246,32 @@ changes:
...
 			yaml: strings.TrimSpace(`
 changes:
 - type: enhancement
   message: New feature has been added
 			`),
 		},
+		{
+			name:     "Next_When_No_Tags_Only_Deps",
+			expected: "v0.1.0",
+			tags:     nil,
```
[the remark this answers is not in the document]
alvarocabanas: No, remember that the bumper takes in consideration the version bumped in the dependency, in this case it passed from 0.0.1 to 0.1.0

src/bumper/bumper.go
alvarocabanas: Modified in last commit

src/bumper/bumper.go
alvarocabanas: Modified

src/app/nextversion/nextversion_test.go
```diff
@@ -239,18 +246,32 @@ changes:
...
 			args:     "--tag-prefix chart- --output-prefix=",
 			yaml: strings.TrimSpace(`
 changes:
 - type: enhancement
   message: New feature has been added
 			`),
 		},
+		{
+			name:     "Next_When_No_Tags_Only_Deps",
+			expected: "v0.1.0",
```
roobre, line 259: I still don't agree with this test case. I think that if there are no versions altogether on the repository, the output should _unconditionally_ be `0.0.1`

alvarocabanas: Even if the depencency bump is a Minor?

alvarocabanas: The thing is in this case, the value from the first version will be calculated between 0.0.1 and the actual bump, and since in this case the dependencies create a minor, that's why it's 0.1.0.
If there would be a Breaking, for example (even if there are no previous releases), the resulting version would be 1.0.0
So, do you want that, whatever is the BumpType, if there are no previous tags, the version resulting to be 0.0.1?

src/git/tag_source.go
```diff
@@ -9,6 +10,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var ErrNoReleases = errors.New("no releases found")
```
roobre, line 13: Seems like we're introducing a new term ("releases") to this object, while we previously referred to either `Tags` or `Versions`. I think the last two make sense (`Version` is an rt concept, and `Tags` is the git concepts where `Versions` are derived). I'd avoid introducing "releases"
```suggestion
var ErrNoReleases = errors.New("no version tags found")
```

roobre (approved):

roobre:
One last question, do we have an integration test (i.e. running the command) capturing this case?

src/git/tag_source.go
alvarocabanas: You are right

PR #96, opened 2022-10-04
Compute also old valid version with brackets

alvarocabanas:
Old valid keepachangelog versions like this one:

````
## [v1.2.3] - 20YY-DD-MM

### Enhancements
- This is in the past and should be preserved
````

Are being removed or not computed.

### DIFF
--- a/src/changelog/sources/markdown/merger/merger.go
+++ b/src/changelog/sources/markdown/merger/merger.go
@@ -43,7 +43,7 @@ func New(ch *changelog.Changelog, newVersion *semver.Version) Merger {
 var (
 	unreleasedHeader = regexp.MustCompile(`(?i)^##\s*unreleased`)
 	heldHeader       = regexp.MustCompile(`(?i)^##\s*held`)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/changelog/sources/markdown/merger/merger.go [hunk 1]
```diff
@@ -43,7 +43,7 @@ func New(ch *changelog.Changelog, newVersion *semver.Version) Merger {
 var (
 	unreleasedHeader = regexp.MustCompile(`(?i)^##\s*unreleased`)
 	heldHeader       = regexp.MustCompile(`(?i)^##\s*held`)
-	l2Header         = regexp.MustCompile(`^##\s*\w`)
+	l2Header         = regexp.MustCompile(`^##\s*[\w\[]`)
```
roobre, line 46: I think essentially any character should work. In fact I think that was my original intention. What about
```suggestion
	l2Header         = regexp.MustCompile(`^##\s*\S`)
```

roobre (approved):
LGTM, good work catching this!

PR #99, opened 2022-10-04
enable go install 

gsanchezgavier:
Currently is not possible to install the `rt` with a `go install` command. This PR address that.


alvarocabanas (approved):
LGTM

PR #102, opened 2022-10-06
feat(next-tag): making output tag configurable

paologallinaharbur:
In some scenarios, you do not want a prefix in the output or you want a different one.

For example, releasing a chart you do not need the tool to return a value such as "v1.y.z" but you would like something like "chart-x-y-z". Moreover, when writing docs you might need no prefix at all "x.y.z"

The old behaviour is kept thanks to the default `v`

### DIFF
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -17,10 +17,11 @@ import (
 )
 
 const (
-	tagPrefix   = "tag-prefix"
-	currentFlag = "current"
-	nextFlag    = "next"
-	gitRootFlag = "git-root"
+	outputPrefix = "output-prefix"
+	tagPrefix    = "tag-prefix"
+	currentFlag  = "current"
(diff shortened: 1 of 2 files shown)


src/app/nextversion/nextversion.go
```diff
@@ -17,10 +17,11 @@ import (
 )
 
 const (
-	tagPrefix   = "tag-prefix"
-	currentFlag = "current"
-	nextFlag    = "next"
-	gitRootFlag = "git-root"
+	outputPrefix = "output-prefix"
```
paologallinaharbur, line 20: name improvement are welcomed 😄

gsanchezgavier (approved):
LGTM

src/app/nextversion/nextversion.go
gsanchezgavier: since the command is next-version , i like this proposed name.

PR #105, opened 2022-10-10
Add all documentation to readme

alvarocabanas:
- Add general README.md documentation.
- Add README.md for each action.


roobre (commented):
I have some opinionated takes on this, so I think the best way to move forward is for me to make changes on top of the commits that are already here, and then we can discuss if we like them or not :)

Some things I think we should change (and I'll try to address):
- I like the diagram, but it is very complete and with a lot of details without any text to introduce it. I think it can become a bit overwhelming. I'll add some paragraphs (and maybe a simpler version) before it.
- Documentation for commands is not autogenerated, and I think it should be. `urfave/cli` has some capabilities we might want to leverage: https://github.com/urfave/cli/blob/77fb75adf7576f5b085b413452f1dc587ef6a609/docs.go
-

alvarocabanas:
> I have some opinionated takes on this, so I think the best way to move forward is for me to make changes on top of the commits that are already here, and then we can discuss if we like them or not :)
> 
> Some things I think we should change (and I'll try to address):
> 
> * I like the diagram, but it is very complete and with a lot of details without any text to introduce it. I think it can become a bit overwhelming. I'll add some paragraphs (and maybe a simpler version) before it.
> * Documentation for commands is not autogenerated, and I think it should be. `urfave/cli` has some capabilities we might want to leverage: https://github.com/urfave/cli/blob/77fb75adf7576f5b085b413452f1dc587ef6a609/docs.go

Sure, go on

README.md
```diff
@@ -1,26 +1,341 @@
...
+
+## Actions
+
+- [Generate YAML changelog](./generate-yaml/README.md)
+- [Is Held](./is-held/README.md)
+- [Link dependencies](./link-dependencies/README.md)
+- [Next Version](./next-version/README.md)
+- [render](./render/README.md)
+- [Update markdown](./update-markdown/README.md)
+- [Validate markdown](./validate-markdown/README.md)
```
roobre, line 329: Will these links work regardless of the page where they're being visited?

I think that if they are on the front page, they will redirect to `https://github.com/newrelic/release-toolkit/generate-yaml/README.md`, which does not exist. The correct URL needs to include `/blob/main/`.

Including `blob/main` unconditionally, however, will not work if you are viewing the README from its dedicated page: https://github.com/newrelic/release-toolkit/blob/main/README.md.

README.md
```diff
@@ -1,26 +1,341 @@
...
+
+## Contributing
+
+Standard policy and procedure across the New Relic GitHub organization.
+
+#### Useful Links
+* [Code of Conduct](./CODE_OF_CONDUCT.md)
+* [Security Policy](./SECURITY.md)
+* [License](./LICENSE)
```
roobre, line 338: As above, links are dangerous.

README.md
```diff
@@ -1,26 +1,341 @@
...
+    message: Support has been removed
+```
+
+This flag can be readily checked using the `is-held` action/command. `is-held` will echo the value of this property to `stdout`, and can be configured to exit with non-zero if `held` is set to `true`. While running as an action, it will also set the `is-held` step output to the value of the property. This allows your pipeline to react to this property and skip the release process if a `## Held` entry is present in your `CHANGELOG.md`.
+
+`update-changelog` will not behave any different in presence of a `## Held` header, and will remove it before integrating the changelog for the latest release.
+
+### Full example of automated releasing pipeline
+
+![release-toolkit.png](release-toolkit.png)
```
roobre, line 206: Will this hyperlink work on the main page of the repo? I don't think it will.

is-held/README.md
```diff
@@ -0,0 +1,51 @@
...
+We encourage you to bring your experiences and questions to the [Explorers Hub](https://discuss.newrelic.com) where our community members collaborate on solutions and new ideas.
+
+## License
+
+release-toolkit is licensed under the [Apache 2.0](http://apache.org/licenses/LICENSE-2.0.txt) License.
+
+## Disclaimer
+
+This tool is provided by New Relic AS IS, without warranty of any kind. New Relic does not guarantee that the tool will: not cause any disruption to services or systems; provide results that are complete or 100% accurate; correct or cure any detected vulnerability; or provide specific remediation advice.
+
```
roobre, line 51: I think all these boilerplatey entries only belong in the main readme, not in the individual ones.

README.md
```diff
@@ -1,26 +1,341 @@
...
+
+The release toolkit is intended to read entries from a CHANGELOG.md following the [keep a changelog format](https://keepachangelog.com/). 
+
+Any change or note, the user wants to add to the next release, should be placed under the Level2 header `Unreleased`.
+
+The `generate-yaml` command will read the entries in the `Unreleased` section, adding all the correct entries to the changelog.yaml file. We describe the types of entries in the next section.
+
+
+
+## CLI
```
roobre, line 229: This readme is very lengthy already, I wonder if we should move all this to a separate `.md` file for clarity. E.g. `README_cli.md`.

generate-yaml/README.md
```diff
@@ -0,0 +1,110 @@
...
+```yaml
+- name: Test generate changelog yaml
+  uses: newrelic/release-toolkit/generate-yaml@v1
+  with:
+    excluded-dirs: charts
+```
+
+## Parameters
+
+### Changes
```
roobre, line 19: @alvarocabanas  I've moved the more detailed section in `README.md` here. I think this is the place where we can elaborate all the nitty-gritty details: Which headers are allowed and which aren't, etc.

README.md
```diff
@@ -1,26 +1,341 @@
...
+   5. Reincorporate all the contents of `changelog.yaml` into `CHANGELOG.md`, under the correct version header.
+
+Each of these steps is performed as a different step in the pipeline, giving maintainers complete flexibility to skip or add intermediate steps. The machine-readable `changelog.yaml` can be easily edited with simple scrips, allowing more flexibility to add, remove, or change changelog entries before the steps that act on them take place.
+
+## Sneak peek
+
+A pipeline using the release toolkit could look like the following:
+
+```yaml
+// TODO
```
roobre, line 26: We should add a short, full example here of how a workflow file looks like. For simplicity we can use `gh release` in the same workflow to illustrate how to compose with `next-version` and `render-markdown`, and `docker build`/`docker push` as simple examples of where the section that builds/publishes the code would live.

README.md
```diff
@@ -1,26 +1,341 @@
...
+This flag can be readily checked using the `is-held` action/command. `is-held` will echo the value of this property to `stdout`, and can be configured to exit with non-zero if `held` is set to `true`. While running as an action, it will also set the `is-held` step output to the value of the property. This allows your pipeline to react to this property and skip the release process if a `## Held` entry is present in your `CHANGELOG.md`.
+
+`update-changelog` will not behave any different in presence of a `## Held` header, and will remove it before integrating the changelog for the latest release.
+
+### Full example of automated releasing pipeline
+
+![release-toolkit.png](release-toolkit.png)
+
+```yaml
+// TODO
```
roobre, line 209: Here we can put a longer example featuring the cron release job.

README.md
```diff
@@ -1,26 +1,341 @@
...
+
+## User-facing changelog
+
+The release toolkit is intended to read entries from a CHANGELOG.md following the [keep a changelog format](https://keepachangelog.com/). 
+
+Any change or note, the user wants to add to the next release, should be placed under the Level2 header `Unreleased`.
+
+The `generate-yaml` command will read the entries in the `Unreleased` section, adding all the correct entries to the changelog.yaml file. We describe the types of entries in the next section.
+
+
```
roobre, line 227: I didn't add a section about `validate-changelog`, I think we should include one under this `Advanced` section.

roobre (changes requested):

README.md
```diff
@@ -1,26 +1,270 @@
...
+Each of these steps is performed as a different step in the pipeline, giving maintainers complete flexibility to skip or add intermediate steps. The machine-readable `changelog.yaml` can be easily edited with simple scripts, allowing more flexibility to add, remove, or change changelog entries before the steps that act on them take place.
 
-RT can be installed with Go by executing the following command:
+## Sneak peek
 
-```shell
-go install github.com/newrelic/release-toolkit@latest
+A pipeline using the release toolkit could look like the following:
+
+```yaml
```
roobre, line 25: I've added an example yaml here, but at this point I think you have more expertise than I do on this :)

LMK if it makes sense to you @alvarocabanas @paologallinaharbur @gsanchezgavier

README.md
```diff
@@ -1,26 +1,270 @@
...
+- `bugfix`: Fixes to incorrect behavior of the application.
+
+Additionally, a section with changes to dependencies is also included after the list of changes.
+
+The Markdown version of the `changelog.yaml` file mentioned above would be:
+
+```markdown
+## v1.2.3 - 2022-10-11
+
+### Important announcement (note)
```
gsanchezgavier, line 128: how this important note is added? Is there an entry type for this ?

validate-markdown/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Parameters
+
+All parameters are optional and match the ones used for the cli command flag, you can see the values and the defaults in [here](../README.md#validate-markdown))
+
+## Outputs
+
+`valid`: Returns `true` if the changelog is valid
+
+## Contributing
```
sigilioso, line 38: The _contributing_, _support_ ..., are included only in the "root README" and the one in `validate-markdown` for any particular reason?

README.md
sigilioso: It seems to be working, at least, when the file is _pre-viewed_ 

image

link-dependencies/README.md
```diff
@@ -0,0 +1,27 @@
...
+  uses: newrelic/release-toolkit/link-dependencies@v1
+  with:
+    dictionary: .github/dictionary.yaml
+```
+
+Dictionary file:
+
+```yaml
+dictionary:
+  common-library: "https://github.com/newrelic/helm-charts/releases/tag/common-library-{{.To}}"
```
gsanchezgavier, line 21: is it To or To.Original?

README.md
roobre: I remember that for a long time, images in `README.md` wouldn't work when the readme is viewed from the repo root (`https://github.com/newrelic/release-toolkit`) because they would generate an image source like:
`https://github.com/newrelic/release-toolkit/release-toolkit.png`
Which Github does not handle well.

Maybe Github has fixed this 🤔

link-dependencies/README.md
alvarocabanas: It depends on each case, because some github repos have v1.1.1 and others have 1.1.1 the original will respect the v in case they have it.

README.md
alvarocabanas: Yes it does work.

README.md
alvarocabanas: I've checked and added bellow where it needs to be fixed.

README.md
```diff
@@ -1,26 +1,270 @@
...
+  - name: Figure out next version automatically
+    id: next-version
+    uses: newrelic/release-toolkit/next-version@v1
+  - name: Generate release notes
+    uses: newrelic/release-toolkit/render-markdown@v1
+    # CHANGELOG.partial.md now contains release notes for this version.
+  - name: Update CHANGELOG.md
+    uses: newrelic/release-toolkit/update-markdown@v1
+    with:
+      next-version: ${{ steps.next-version.outputs.next-version }}
```
alvarocabanas, line 51: this should be version without next-

README.md
```diff
@@ -1,26 +1,270 @@
...
+    # - Entries generated from bot commits (e.g. dependabot/renovate)
+  # - name: You can hack changelog.yaml!
+  #   run: |
+  #     yq ...
+  #     custom-script.sh ...
+  - name: Figure out next version automatically
+    id: next-version
+    uses: newrelic/release-toolkit/next-version@v1
+  - name: Generate release notes
+    uses: newrelic/release-toolkit/render-markdown@v1
```
alvarocabanas, line 46: the action is called newrelic/release-toolkit/render@v1

README.md
alvarocabanas: This will be added as an additional note exactly as it is

validate-markdown/README.md
alvarocabanas: mmm.. good point, we should add it to all or only to the main readme at least

PR #107, opened 2022-10-10
Link dependencies fix removed meta

alvarocabanas:
- This PR removes a bug found on link-dependency that was caused by a missing field in the dependency entryMeta unmarshal.


paologallinaharbur (approved):

roobre (approved):
Looking awesome, thanks for catching this!

PR #112, opened 2022-10-11
NextVersion error if no new version

alvarocabanas:
- Exit 1 and show error if nextVersion detects there is not a new version to release.

### DIFF
--- a/next-version/action.yml
+++ b/next-version/action.yml
@@ -17,6 +17,10 @@ inputs:
     description: The prefix to prepend when printing the output version
     required: false
     default: "v"
+  fail:
+    description: Fail if no new version found, by default the current version will be returned in that case
+    required: false
+    default: "0"
 outputs:
   next-version:
     description: Semver next version, with leading v
@@ -33,3 +37,4 @@ runs:
     - ${{ inputs.tag-prefix }}
     - --output-prefix
     - ${{ inputs.output-prefix }}
+    - --fail=${{ inputs.fail }}
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -26,6 +26,7 @@ const (
 	gitRootFlag       = "git-root"
 	BumpCapFlag       = "bump-cap"
 	DependencyCapFlag = "dependency-cap"
+	failFlag          = "fail"
 )
 
 const nextVersionOutput = "next-version"
@@ -83,6 +84,13 @@ next-version will exit with an error if no previous versions are found in the gi
 			EnvVars: common.EnvFor(DependencyCapFlag),
 			Usage:   "In case of having to bump the version of base on a dependency, limit to this semVer type",
 			Value:   string(bump.MajorName),
+		},
+		&cli.BoolFlag{
+			Name:    failFlag,
+			EnvVars: common.EnvFor(failFlag),
+			Usage: "If set, command will exit with a code of 1 if no new version bump is produced. If not set," +
+				"the current version will be returned.",
+			Value: false,
 		},
 	},
 	Action: NextVersion,
@@ -154,6 +162,13 @@ func NextVersion(cCtx *cli.Context) error {
 
 	case nextOverride == nil && next != nil:
 		nextStr = next.String()
+		// If we don't have an override, the bumper did not bump anything, and the user has set `--fail`, then we should error out.
+		if errors.Is(err, bumper.ErrNoNewVersion) {
+			log.Warnf("None of the changelog entries produced a version bump, returning current version")
+			if cCtx.Bool("fail") {
+				return fmt.Errorf("failing by user request: %w", err)
+			}
+		}
 
 	case nextOverride == nil && next == nil:
 		return fmt.Errorf("bumping source: %w", err)
--- a/src/app/nextversion/nextversion_test.go
+++ b/src/app/nextversion/nextversion_test.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/newrelic/release-toolkit/src/app"
 	"github.com/newrelic/release-toolkit/src/bump"
-	"github.com/newrelic/release-toolkit/src/bumper"
 )
 
 //nolint:paralleltest, funlen // urfave/cli cannot be tested concurrently.
@@ -88,9 +87,9 @@ changes:
 			`),
 		},
 		{
-			name:          "No_Bump",
-			args:          "-current v1.2.3",
-			errorExpected: bumper.ErrNoNewVersion,
+			name:     "No_Bump",
+			args:     "-current v1.2.3",
+			expected: "v1.2.3",
 			yaml: strings.TrimSpace(`
 changes: []
 			`),
@@ -329,6 +328,15 @@ dependencies:
   to: 1.0.0
 			`),
 		},
+		{
+			name:     "When_Repo_Has_No_Canges_But_Fail_Is_False",
+			expected: "v0.1.0",
+			args:     "--fail=0",
+			tags:     []string{"v0.1.0"},
+			yaml: strings.TrimSpace(`
+notes: "adfafds"
+			`),
+		},
 	} {
 		tc := tc
 		//nolint:paralleltest // urfave/cli cannot be tested concurrently.
@@ -407,6 +415,17 @@ dependencies:
 - name: foobar
   from: 0.0.1
   to: 0.1.0
+			`),
+		},
+		{
+			name: "When_Repo_Has_No_Changes_And_Must_Fail",
+			args: "-fail=1",
+			createRepoFunc: func(t *testing.T) string {
+				t.Helper()
+				return repoWithTags(t, "v0.1.0") // Empty tag list.
+			},
+			yaml: strings.TrimSpace(`
+notes: "adfafds"
 			`),
 		},
 	} {
--- a/src/app/nextversion/nextversion_test.go
+++ b/src/app/nextversion/nextversion_test.go
@@ -11,16 +11,18 @@ import (
 
 	"github.com/newrelic/release-toolkit/src/app"
 	"github.com/newrelic/release-toolkit/src/bump"
+	"github.com/newrelic/release-toolkit/src/bumper"
 )
 
-//nolint:paralleltest // urfave/cli cannot be tested concurrently.
+//nolint:paralleltest, funlen // urfave/cli cannot be tested concurrently.
 func TestNextVersion_Without_Repo(t *testing.T) {
 	for _, tc := range []struct {
-		name       string
-		yaml       string
-		expected   string
-		args       string
-		globalargs string
+		name          string
+		yaml          string
+		expected      string
+		errorExpected error
+		args          string
+		globalargs    string
 	}{
 		{
 			name:     "Overrides_Next_And_Current",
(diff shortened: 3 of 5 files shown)


src/bump/type.go
```diff
@@ -53,9 +59,9 @@ func From(previous, current *semver.Version) Type {
 }
 
 // Bump returns a new version after bumping it according to the specified bump bump.
-func Bump(version *semver.Version, bt Type) *semver.Version {
+func Bump(version *semver.Version, bt Type) (*semver.Version, error) {
```
roobre, line 62: I agree with the goal of make `next-version` fail, but I think this is a responsibility of the command. Could we make the command fail if `version == Bump(version, None)`, and not return any error here?

roobre (changes requested):

roobre:
If we want to give users a way to gracefully stop CI/CD pipelines if there is nothing to release, we might want to consider adding an `is-empty` command.

Checking if `next-version` errored as an indirect means of checking if there is anything to release doesn't feel right to me.

src/bump/type.go
alvarocabanas: I've moved this error logic in the last commit to the bumper.BumpSource, this method is already erroring when there are no tags, and I find it correct, because you are trying to bump but the source doesn't have anything to bump. 
If you still think this is better responsibility of the command, then this :
```
	sort.Slice(versions, func(i, j int) bool {
		// We use GreaterThan as lessFunc to sort from largest to smallest.
		return versions[i].GreaterThan(versions[j])
	})
```
needs to be duplicated in the command in order to know what is the current version in the source and then make the comparison.

alvarocabanas:
> If we want to give users a way to gracefully stop CI/CD pipelines if there is nothing to release, we might want to consider adding an `is-empty` command.
> 
> Checking if `next-version` errored as an indirect means of checking if there is anything to release doesn't feel right to me.
@roobre, I see your point, but at the same time, I also expect a next-version to give me a next-version or tell me that there is none.

src/app/nextversion/nextversion_test.go
```diff
@@ -105,11 +115,11 @@ changes:
 			app.Writer = buf
 
 			err = app.Run(strings.Fields(fmt.Sprintf("rt -yaml %s %s next-version %s", yamlPath, tc.globalargs, tc.args)))
-			if err != nil {
-				t.Fatalf("Error running app: %v", err)
+			if !errors.Is(err, tc.errorExpected) {
+				t.Fatalf("Expected %v, got %v", tc.errorExpected, err)
```
roobre, line 119:
```suggestion
				t.Fatalf("Expected error %v, got %v", tc.errorExpected, err)
```

src/app/nextversion/nextversion_test.go
```diff
@@ -105,11 +115,11 @@ changes:
...
 			}
 
-			if actual := buf.String(); actual != tc.expected+"\n" {
+			if actual := buf.String(); tc.expected != "" && actual != tc.expected+"\n" {
```
roobre, line 122: Can we make this a different `if` earlier? Human brains suck reading nested conditions.

src/bumper/bumper.go
```diff
@@ -10,6 +11,8 @@ import (
 	"github.com/newrelic/release-toolkit/src/version"
 )
 
+var ErrNoNewVersion = errors.New("no newer version to be bumped")
```
roobre, line 14: I'm still not convinced about the erroring logic being in the bumper. I think the action we want to take is fundamentally a user-facing one, so I think it may be better for this logic to be in the command instead.

roobre (commented):
After thinking about it, it is true that `next-version` should provide some kind of feedback if there is no next version.

However I still think we should protect users from using this as a check for releasability. I've left a few minor comments but, in addition, I would like to suggest the following approach:

1. Make the `next-version` command print a warning if the version bump is none
2. Add a flag (e.g. `--fail`, or `--exit-code`) *defaulting to false* for the command to exit with non-zero code if this is the case.

WDYT about this approach?

src/bumper/bumper.go
alvarocabanas: But if we do that, we will need the command to iterate all the tags to get the first. The command doesn't know nothing about the current version.

alvarocabanas: And I see it here, I expect the BumpSource command to Bump the source or tell me it couldn't.

src/app/nextversion/nextversion_test.go
alvarocabanas: done

alvarocabanas:
> After thinking about it, it is true that `next-version` should provide some kind of feedback if there is no next version.
> 
> However I still think we should protect users from using this as a check for releasability. I've left a few minor comments but, in addition, I would like to suggest the following approach:
> 
> 1. Make the `next-version` command print a warning if the version bump is none
> 2. Add a flag (e.g. `--fail`, or `--exit-code`) _defaulting to false_ for the command to exit with non-zero code if this is the case.
> 
> WDYT about this approach?

Yes, sure, makes sense.

src/bumper/bumper.go
roobre: > But if we do that, we will need the command to iterate all the tags to get the last. The command doesn't know nothing about the current version.

When I wrote this I had, for some reason, the wrong idea that the command was dealing with `bump.Type` directly. This is in fact not true, so your comment is on point. `bump.Bump` should not return an error (just `bump.None` if there is none), but `bumper.Bump` should return an error so the command can catch it.

roobre:
>  I am not completely sure about this behavior.

My reasoning regarding this is the same as exposed here: https://github.com/newrelic/release-toolkit/pull/112#issuecomment-1274955304.
There are actual erros that can cause `next-version` to fail, mainly due to some errors extacting tags from the git repo. I think that if we make it error also when there is no next version, users will immediately think they can use `next-version` errors to tell there is no release to be made, and make the pipeline exit cleanly. This may not be the case if `next-version` is erroring for any of the reasons explained above (something wrong with the repo), so I think we're training users to do something dangerous here.

I think it is very reasonable for `next-version` to output a human-readable error when this is the case. Something on the lines of:

```$ rt next-version
Warning: None of the changelog entries produced a version bump, returning current version
v1.2.3
$ echo $?
0
```

alvarocabanas:
> Warning: None of the changelog entries produced a version bump, returning current version

I've added this behavior you comment to the last commit https://github.com/newrelic/release-toolkit/pull/112/commits/899c97ec495168a70e13c9703da1c7bda66281cc, tell me WDYT

next-version/action.yml
```diff
@@ -33,3 +37,5 @@ runs:
     - ${{ inputs.tag-prefix }}
     - --output-prefix
     - ${{ inputs.output-prefix }}
+    - --fail
+    - ${{ inputs.fail }}
```
roobre, line 41: I might be remembering incorrecty, but I think this syntax will not work for `BoolFlags` (urfave/cli expects you to use `--fail=0`). Can we add an integration test covering this flag to confirm it?

src/app/nextversion/nextversion.go
```diff
@@ -146,6 +154,12 @@ func NextVersion(cCtx *cli.Context) error {
 
 	case nextOverride == nil && next != nil:
 		nextStr = next.String()
+		if err != nil {
+			if cCtx.Bool("fail") {
+				return cli.Exit(fmt.Errorf("bumping source: %w", err), 1)
+			}
+			log.Warnf("None of the changelog entries produced a version bump, returning current version")
+		}
```
roobre, line 162: I think nesting this much makes the code a bit unreadable (mostly because of this big ugly switch case I introduced 😓)

We should probably refactor it at this point. For now, could we make this error check specific `errors.Is` and add a comment for our future selves?

```suggestion
		// If we don't have an override, the bumper did not bump anything, and the user has set `--fail`, then we should error out.
		if errors.Is(err, ErrNoNewVersion)  {
			log.Warnf("None of the changelog entries produced a version bump, returning current version")
			if cCtx.Bool("fail") {
				return cli.Exit(fmt.Errorf("failing by user request: %w", err), 1)
			}
		}
```

src/bumper/bumper.go
roobre, line 14:
```suggestion
var ErrNoNewVersion = errors.New("bump did not change the version")
```

next-version/action.yml
alvarocabanas: sure

alvarocabanas: Fixed in last commit and test added.

src/app/nextversion/nextversion.go
```diff
@@ -154,6 +162,14 @@ func NextVersion(cCtx *cli.Context) error {
 
 	case nextOverride == nil && next != nil:
 		nextStr = next.String()
+		// If we don't have an override, the bumper did not bump anything, and the user has set `--fail`, then we should error out.
+		if errors.Is(err, bumper.ErrNoNewVersion) {
+			log.Warnf("None of the changelog entries produced a version bump, returning current version")
+			a := cCtx.Bool("fail")
+			if a {
```
roobre, line 169: This `a` looks like a debugging leftover 😉

src/app/nextversion/nextversion.go
```diff
@@ -84,6 +85,13 @@ next-version will exit with an error if no previous versions are found in the gi
 			Usage:   "In case of having to bump the version of base on a dependency, limit to this semVer type",
 			Value:   string(bump.MajorName),
 		},
+		&cli.BoolFlag{
+			Name:    failFlag,
+			EnvVars: common.EnvFor(failFlag),
+			Usage: "If set, command will exit with a code of 1 if no new version bump is produced. If not set," +
+				"the current version will be returned in that case.",
```
roobre, line 92: Minor wording nit
```suggestion
				"the current version will be returned.",
```

src/app/nextversion/nextversion.go
alvarocabanas: yes...

src/app/nextversion/nextversion_test.go
```diff
@@ -85,6 +86,14 @@ changes:
...
 			`),
 		},
+		{
+			name:     "No_Bump",
+			args:     "-current v1.2.3",
+			expected: "v1.2.3",
+			yaml: strings.TrimSpace(`
+changes: []
+			`),
+		},
```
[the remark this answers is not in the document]
alvarocabanas: It's not possible, because if it fails the test exits, it's not possible to test an os.exit.

src/app/nextversion/nextversion.go
alvarocabanas: fixed

src/app/nextversion/nextversion_test.go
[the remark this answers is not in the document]
alvarocabanas: You are right, I don't know why I did it with the cli.Exit when it's better as you suggest.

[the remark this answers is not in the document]
alvarocabanas: Fixed.

src/app/nextversion/nextversion_test.go
```diff
@@ -398,6 +417,17 @@ dependencies:
   to: 0.1.0
 			`),
 		},
+		{
+			name: "When_Repo_Has_No_Changes_And_Must_Fail",
+			args: "-fail=1",
+			createRepoFunc: func(t *testing.T) string {
+				t.Helper()
+				return repoWithTags(t, "v0.1.0") // Empty tag list.
```
roobre, line 425: Not actually empty
```suggestion
				return repoWithTags(t, "v0.1.0")
```

roobre (approved):

Issue #109, opened 2022-10-10
The tools is considering tag coming from different branches and remotes


paologallinaharbur:
It seems that locally the tools takes into considerations tags created in different remotes and following a different history.

At least for dependabot:

<img width="1123" alt="Screenshot 2022-10-10 at 18 45 29" src="https://user-images.githubusercontent.com/43335750/194915456-53a2845a-15be-4f40-94e0-49b0ba37a819.png">

Here it skips releases using as the latest the one coming from a fork

paologallinaharbur:
Moreover, since an off-branch tag is selected then all commits are retrieved since such tags is never found in the list of commits by:
```
func (s *RepoCommitsGetter) Commits(lastHash string) ([]Commit, error) {
```

for that reason in the image above an old commit from Guillermo was selected

paologallinaharbur:
```
All tags available [{"newrelic-prometheus-agent-0.0.4" "f4cde03bc780d8aae9f7ced8434fc5cd14e13b33"} {"newrelic-prometheus-agent-0.0.5" "25153bf1101695b03f8a572b395f0070d39387fd"} {"newrelic-prometheus-agent-0.0.6" "fadf3d63256e67162dba09be179483a37ab57c22"} {"newrelic-prometheus-agent-0.0.7" "3dffb30d2cd0ec323aef8aee5cef69e661385f0f"} {"newrelic-prometheus-agent-0.1.0" "82216e252f3bb842780b7605097857036d2b5692"}] 
Latest tag 82216e252f3bb842780b7605097857036d2b5692      
Number of commits  since latest tag 133                                          
```

roobre:
Agreed, I think only tags reachable from the currently checked-out branch should be considered for this.

In the future we might want to add a flag to tweak this behavior, but reachable from `HEAD` is more than enough for now.

PR #113, opened 2022-10-11
[NR-57250] fix(tags): ignoring tags from different branches

paologallinaharbur:
Fix https://github.com/newrelic/release-toolkit/issues/109

The approach is quite simple, we check all tags and discard the ones not found in the list of commits since the root one (following the history).

```
rt validate-markdown --markdown "charts/newrelic-prometheus-agent"/CHANGELOG.md
rt generate-yaml --markdown "charts/newrelic-prometheus-agent"/CHANGELOG.md --included-dirs "charts/newrelic-prometheus-agent"  --tag-prefix newrelic-prometheus-agent-
INFO[0000] Ignoring newrelic-prometheus-agent-0.0.6 since it belongs to a different branch 
INFO[0000] Ignoring newrelic-prometheus-agent-0.0.7 since it belongs to a different branch 
INFO[0000] Ignoring newrelic-prometheus-agent-0.1.0 since it belongs to a different branch 
INFO[0000] Renovate source did not find any commit since "25153bf1101695b03f8a572b395f0070d39387fd" 
```
Screenshot 2022-10-11 at 16 56 06

### DIFF
--- a/src/app/generate/generate.go
+++ b/src/app/generate/generate.go
@@ -191,13 +191,15 @@ func addDepSource(cCtx *cli.Context, sources []changelog.Source, includedDirs, e
 }
 
 func tagVersionGetter(cCtx *cli.Context) (*git.TagsSource, error) {
-	var tagOpts []git.TagOptionFunc
+	workDir := cCtx.String(gitRootFlag)
+	commitsGetter := git.NewRepoCommitsGetter(workDir)
+
+	tagOpts := []git.TagOptionFunc{git.TagsMatchingCommits(commitsGetter)}
 	if matching := cCtx.String(tagPrefixFlag); matching != "" {
-		tagOpts = append(tagOpts, git.TagsMatching("^"+matching))
+		tagOpts = append(tagOpts, git.TagsMatchingRegex("^"+matching))
 	}
 
-	workDir := cCtx.String(gitRootFlag)
-	src, err := git.NewRepoTagsGetter(workDir, git.NewRepoCommitsGetter(workDir), tagOpts...)
+	src, err := git.NewRepoTagsGetter(workDir, tagOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("creating source for git tags: %w", err)
 	}
--- a/src/app/generate/generate.go
+++ b/src/app/generate/generate.go
@@ -196,7 +196,8 @@ func tagVersionGetter(cCtx *cli.Context) (*git.TagsSource, error) {
 		tagOpts = append(tagOpts, git.TagsMatching("^"+matching))
 	}
 
-	src, err := git.NewRepoTagsGetter(cCtx.String(gitRootFlag), tagOpts...)
+	workDir := cCtx.String(gitRootFlag)
+	src, err := git.NewRepoTagsGetter(workDir, git.NewRepoCommitsGetter(workDir), tagOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("creating source for git tags: %w", err)
 	}
--- a/src/app/nextversion/nextversion_test.go
+++ b/src/app/nextversion/nextversion_test.go
@@ -21,8 +21,8 @@ func TestNextVersion_Without_Repo(t *testing.T) {
 		globalargs string
 	}{
 		{
-			name:     "Overrides_Next",
-			args:     "-next v0.0.1",
+			name:     "Overrides_Next_And_Current",
+			args:     "-next v0.0.1 -current v1.2.3",
 			expected: "v0.0.1",
 			yaml: strings.TrimSpace(`
 notes: |-
@@ -157,7 +157,8 @@ dependencies:
 		},
 		{
 			name:     "Overrides_Current",
-			expected: "v3.0.0",
+			args:     "--current v0.0.1",
+			expected: "v1.0.0",
 			yaml: strings.TrimSpace(`
 notes: |-
     ### Important announcement (note)
--- a/src/git/tag.go
+++ b/src/git/tag.go
@@ -39,7 +39,7 @@ func TagsMatchingRegex(regex string) TagOptionFunc {
 	}
 }
 
-// TagsMatchingCommits returns an option that will make the getter to ignore tags that do not match commits.
+// TagsMatchingCommits returns an option that will skip tags that do not point to a commit reachable from HEAD.
 func TagsMatchingCommits(getter CommitsGetter) TagOptionFunc {
 	return func(s *RepoTagsGetter) error {
 		branchCommits, err := getter.Commits(EmptyTreeID)
--- a/src/git/tag.go
+++ b/src/git/tag.go
@@ -108,7 +108,7 @@ func (s *RepoTagsGetter) Tags() ([]Tag, error) {
 }
 
 // getCommitsMap retrieves the list of commits from HEAD and put it in a map in order to
-// filter out all the tags not belonging to the current branch
+// filter out all the tags not belonging to the current branch.
 func getCommitsMap(commitGetter CommitsGetter) (map[string]bool, error) {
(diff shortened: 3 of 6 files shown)


src/git/tag.go
```diff
@@ -42,8 +43,9 @@ var MatchAllTags = regexp.MustCompile("")
 
 func NewRepoTagsGetter(workDir string, opts ...TagOptionFunc) (*RepoTagsGetter, error) {
 	s := &RepoTagsGetter{
-		workDir: workDir,
-		match:   MatchAllTags,
+		workDir:       workDir,
+		match:         MatchAllTags,
+		commitsGetter: NewRepoCommitsGetter(workDir),
 	}
```
alvarocabanas, line 49: Since You are composing the RepTagsGetter with a commitsGetter interface, then I would inject the NewRepoCommitsGetter from the call and change the signature to b eNewRepoTagsGetter(workDir string, commitsGetter CommitsGetter, opts ...TagOptionFunc).

alvarocabanas (commented):
I left a small comment, apart from that, it looks good, it's the only approach we can take with go-git, what I would like is to see is the performance in a repo with many commits.

src/git/tag.go
paologallinaharbur: Makes sense!

src/app/generate/generate.go
```diff
@@ -196,7 +196,8 @@ func tagVersionGetter(cCtx *cli.Context) (*git.TagsSource, error) {
 		tagOpts = append(tagOpts, git.TagsMatching("^"+matching))
 	}
 
-	src, err := git.NewRepoTagsGetter(cCtx.String(gitRootFlag), tagOpts...)
+	workDir := cCtx.String(gitRootFlag)
+	src, err := git.NewRepoTagsGetter(workDir, git.NewRepoCommitsGetter(workDir), tagOpts...)
```
roobre, line 200: I'm not a fan of putting a `RepoCommitsGetter` inside the `RepoTagsGetter`, I don' think the TagsGetter needs Commits to perform its basic functionality.

If I understood the code correctly, what we want to do here is to _filter_ tags after we have collected them, right? Meaning, we don't use any information from the tags that is not returned as `[]Tag` by `RepoTagsGetter.Tags`.

If this is the case I think that what we should do here is have a TagsFilterer that can work on what `RepoTagsGetter` returns.

```go
tagGetter := git.NewRepoTagsGetter(workDir, tagOpts...) // Same signature, object, and behavior as before this PR
tags := tagGetter.Tags()
//...
filterer := ReachableTagsFilterer(git.NewRepoCommitsGetter(workDir))
tags = filterer.Filter(tags)
```

src/git/tag.go
```diff
@@ -94,3 +106,19 @@ func (s *RepoTagsGetter) Tags() ([]Tag, error) {
...
+		return nil, fmt.Errorf("getting git commits since empty tree: %w", err)
+	}
+
+	commitMap := map[string]bool{}
+	for _, c := range branchCommits {
+		commitMap[c.Hash] = true
+	}
+
+	return commitMap, nil
+}
```
roobre, line 124: This dangling (i.e. not tied to an object) function is a very good pointer that we should have an object doing this work.

roobre (changes requested):

src/app/generate/generate.go
paologallinaharbur: I assumed that the basic functionality was retrieving the tags from a specific branch.

However, I can extract it, I just fear that in the future someone could use `.Tag`  without knowing that the filter is not applied by default

paologallinaharbur: would make sense to extract as well:
```
match         *regexp.Regexp
[...]
func TagsMatching(regex string) TagOptionFunc
```

paologallinaharbur: Now it follows the approach of the tags regex and it is "pluggable". 
If we want a separate all filtering then I would create something like `NewCommitFilter` for both the regex and the branch

src/git/tag.go
```diff
@@ -19,21 +19,39 @@ type TagsGetter interface {
...
+	matchRegex   *regexp.Regexp
+	matchCommits map[string]bool
 }
 
 type TagOptionFunc func(s *RepoTagsGetter) error
 
-// TagsMatching returns an option that will make the getter to ignore tags that do not match regex.
-func TagsMatching(regex string) TagOptionFunc {
+// TagsMatchingRegex returns an option that will make the getter to ignore tags that do not match regex.
+func TagsMatchingRegex(regex string) TagOptionFunc {
```
roobre, line 30: What is the motivation for this change? The name of the parameter here `regex string` should be already hinting what tags should match (a regex)

src/app/generate/generate.go
roobre: > I see it as a similar functionality to: `func TagsMatching(regex string) TagOptionFunc`

This is a good point. The biggest difference I see is that matching a regex is a very easy thing to do (we don't require any dependency in the object), but filtering from a list of commits it's not, as it requires you to inject the CommitsGetter in the equation.

Adding `func TagsMatchingCommits(source CommitSource) TagOptionFunc` looks fine to me.

Extracting both regex filtering _and_ commit filtering out looks even better. In the end, we also have a tag replacement function, that is already extracted and performed by the `TagsSource`:
https://github.com/newrelic/release-toolkit/blob/e6cbc226f6c96c11d389232cfa7f92ce6774437e/src/git/tag_source.go#L35-L35

src/git/tag.go
paologallinaharbur: If you are in the file `A` and you see a function call:
```
a.TagsMatching("abc")
```

Are you sure what is matching? Is is a string, a prefix or a regex? That was the reasoning behind 😅

src/app/generate/generate.go
paologallinaharbur: Ok, then we can leave it like this and wait for a major refactor when the v1 comes

src/app/nextversion/nextversion_test.go
```diff
@@ -21,8 +21,8 @@ func TestNextVersion_Without_Repo(t *testing.T) {
 		globalargs string
 	}{
 		{
-			name:     "Overrides_Next",
-			args:     "-next v0.0.1",
+			name:     "Overrides_Next_And_Current",
+			args:     "-next v0.0.1 -current v1.2.3",
```
paologallinaharbur, line 25: Without a repo we cannot check current tags and commits

src/git/tag.go
```diff
@@ -19,21 +19,39 @@ type TagsGetter interface {
...
 			return fmt.Errorf("compiling %q: %w", regex, err)
 		}
 
-		s.match = rgx
+		s.matchRegex = rgx
+		return nil
+	}
+}
+
+// TagsMatchingCommits returns an option that will make the getter to ignore tags that do not match commits.
```
roobre, line 42:
```suggestion
// TagsMatchingCommits returns an option that will skip tags that do not point to a commit reachable from HEAD.
```

src/git/tag_source_test.go
```diff
@@ -151,14 +214,16 @@ func TestRepoTagsSource_LastVersionHash(t *testing.T) {
 		{
 			name: "Matching_Leading_v",
 			tagOpts: []git.TagOptionFunc{
-				git.TagsMatching("^v"),
+				git.TagsMatchingRegex("^v"),
+				git.TagsMatchingCommits(commitsGetter),
```
roobre, line 218: Why are we adding the `TagsMatchingCommits` opt on this test? It was passing before and we haven't changed the test requirements.

roobre (approved):
LGTM, left a small suggestion and a question regarding a test change. Good job!

src/git/tag_source_test.go
paologallinaharbur: I restored the old tests, there was no need for it

Issue #108, opened 2022-10-10
Regex for dependabot is too permissive for generate-yaml


paologallinaharbur:
[This](https://github.com/newrelic/newrelic-prometheus-configurator/pull/125/files) commit is interpreted as a dependabot one.

```
notes: ""
changes: []
dependencies:
    - name: configurator
      meta:
        pr: "125"
        commit: 25153bf1101695b03f8a572b395f0070d39387fd
```

paologallinaharbur:
@gsanchezgavier did you see already that issue?

gsanchezgavier:
it looks like also related to this #92 , perhaps the one you created is the source and #92 a side effect.

roobre:
This is something we should fix, most likely by checking the author as we do with renovate 
https://github.com/newrelic/release-toolkit/blob/7402f581e2c56cec1d4db50c88b24ddfb347f089/src/changelog/sources/renovate/source.go#L52

PR #118, opened 2022-10-14
[NR-57248] fix(dependabot): check autor in commit

paologallinaharbur:
Fix https://github.com/newrelic/release-toolkit/issues/108

I am keeping both checks, author, and commit message in case there is no info that can be used

### DIFF
--- a/src/changelog/sources/dependabot/source_test.go
+++ b/src/changelog/sources/dependabot/source_test.go
@@ -40,134 +40,140 @@ func TestSource_Source(t *testing.T) {
 		expected []changelog.Dependency
 	}{
 		{
-			commit: git.Commit{Message: "Whatever actions/github-script from 1.0.2 to 1.0.4."},
+			commit: git.Commit{Message: "Whatever actions/github-script from 1.0.2 to 1.0.4.", Author: "dependabot[bot]"},
 		},
 		{
-			commit: git.Commit{Message: "Non matching"},
+			commit: git.Commit{Message: "Non matching", Author: "dependabot[bot]"},
 		},
 		{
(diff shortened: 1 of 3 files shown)


src/changelog/sources/dependabot/source_test.go
```diff
@@ -40,134 +40,140 @@ func TestSource_Source(t *testing.T) {
 		expected []changelog.Dependency
 	}{
 		{
-			commit: git.Commit{Message: "Whatever actions/github-script from 1.0.2 to 1.0.4."},
+			commit: git.Commit{Message: "Whatever actions/github-script from 1.0.2 to 1.0.4.", Author: "dependabot[bot]"},
```
roobre, line 43: To make the test readable, we can "cheat" a bit and, in the test body, set `Author` to a default value if it's empty.

src/changelog/sources/renovate/source.go
```diff
@@ -47,8 +47,6 @@ func (r Source) Changelog() (*changelog.Changelog, error) {
 		var commitDependencies []changelog.Dependency
 
 		commitLine := strings.Split(c.Message, "\n")[0]
-		// Contrary to dependabot, where we can reliably tell if a message comes from it by just inspecting the message,
```
roobre, line 50: > where we can reliably tell

Ominous foreshadowing 🤣

roobre (approved):
LGTM, left a small suggestion regarding tests

src/changelog/sources/dependabot/source_test.go
paologallinaharbur: done

PR #124, opened 2022-11-29
git/test: specify default branch name on git init

roobre:
Before this commit, tests would fail if they are run on an environment where the user has git configured to use a default branch name that is not `master` (e.g. `main`). This PR fixes that by instructing `git init` to use a predictable name for the deafault branch.


sigilioso (approved):
LGTM

Issue #115, opened 2022-10-11
Expose dependency bump cap and default to Minor


roobre:
Right now, a major bump in a dependency will result in a major bump performed by `next-version`. There is a mechanism implement in-code to prevent this from happening:

https://github.com/newrelic/release-toolkit/blob/7402f581e2c56cec1d4db50c88b24ddfb347f089/src/bumper/bumper.go#L20

But there isn't a way to access this from the CLI/action.

We should expose this behavior.

PR #128, opened 2023-01-23
Add version limit to next-version command (NR-79791)

kang-makes:
Right now, a major bump in a dependency will result in a major bump performed by `next-version`. There is a mechanism implemented to prevent this from happening:

https://github.com/newrelic/release-toolkit/blob/7402f581e2c56cec1d4db50c88b24ddfb347f089/src/bumper/bumper.go#L20

This PR should expose this behavior. Solves #115

## Additional notes for the reviewer:

I saw that there was an issue in the bumper that was capping the dependency version using the value from the entry cap, resulting in the dependency cap being ignored. Not only I fixed it but I also added tests as that option was not being tested at all.

### DIFF
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -3,6 +3,7 @@ package nextversion
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/Masterminds/semver"
 	"github.com/newrelic/release-toolkit/src/app/common"
@@ -115,8 +116,8 @@ func NextVersion(cCtx *cli.Context) error {
 		return err
 	}
 
-	entryCap := bump.NameToType(cCtx.String(limitVersionBumpToFlag))
-	dependencyCap := bump.NameToType(cCtx.String(limitDependencyBumpToFlag))
+	entryCap := bump.NameToType(strings.ToLower(cCtx.String(limitVersionBumpToFlag)))
+	dependencyCap := bump.NameToType(strings.ToLower(cCtx.String(limitDependencyBumpToFlag)))
 
 	bmpr := bumper.New(
 		ch,
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -117,11 +117,11 @@ func NextVersion(cCtx *cli.Context) error {
 
 	entryCap, err := bump.NameToType(cCtx.String(BumpCapFlag))
 	if err != nil {
-		return fmt.Errorf("error with the changes' cap: %w", err)
+		return fmt.Errorf("parsing version bump cap: %w", err)
 	}
 	dependencyCap, err := bump.NameToType(cCtx.String(DependencyCapFlag))
 	if err != nil {
-		return fmt.Errorf("error with dependencies' cap: %w", err)
+		return fmt.Errorf("parsing dependency bump: %w", err)
 	}
 
 	bmpr := bumper.New(ch)
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -119,11 +119,10 @@ func NextVersion(cCtx *cli.Context) error {
 	entryCap := bump.NameToType(strings.ToLower(cCtx.String(BumpCapFlag)))
 	dependencyCap := bump.NameToType(strings.ToLower(cCtx.String(DependencyCapFlag)))
 
-	bmpr := bumper.New(
-		ch,
-		bumper.WithEntryCap(entryCap),
-		bumper.WithDependencyCap(dependencyCap),
-	)
+	bmpr := bumper.New(ch)
+	bmpr.EntryCap = entryCap
+	bmpr.DependencyCap = dependencyCap
+
 	next, err := bmpr.BumpSource(versionSrc)
 
 	nextStr := ""
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -7,6 +7,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/newrelic/release-toolkit/src/app/common"
 	"github.com/newrelic/release-toolkit/src/app/gha"
+	"github.com/newrelic/release-toolkit/src/bump"
 	"github.com/newrelic/release-toolkit/src/bumper"
 	"github.com/newrelic/release-toolkit/src/changelog"
 	"github.com/newrelic/release-toolkit/src/git"
@@ -17,11 +18,13 @@ import (
 )
 
 const (
-	outputPrefix = "output-prefix"
-	tagPrefix    = "tag-prefix"
-	currentFlag  = "current"
-	nextFlag     = "next"
-	gitRootFlag  = "git-root"
+	outputPrefix              = "output-prefix"
+	tagPrefix                 = "tag-prefix"
+	currentFlag               = "current"
+	nextFlag                  = "next"
+	gitRootFlag               = "git-root"
+	limitVersionBumpToFlag    = "limit-version-bump-to"
+	limitDependencyBumpToFlag = "limit-dependency-bump-to"
 )
 
 const nextVersionOutput = "next-version"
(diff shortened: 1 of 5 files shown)


src/app/nextversion/nextversion.go
```diff
@@ -17,11 +19,13 @@ import (
...
-	currentFlag  = "current"
-	nextFlag     = "next"
-	gitRootFlag  = "git-root"
+	outputPrefix              = "output-prefix"
+	tagPrefix                 = "tag-prefix"
+	currentFlag               = "current"
+	nextFlag                  = "next"
+	gitRootFlag               = "git-root"
+	limitVersionBumpToFlag    = "limit-version-bump-to"
+	limitDependencyBumpToFlag = "limit-dependency-bump-to"
```
roobre, line 28: These seem a bit long arg names. Could we pick a bit shorter ones?
```suggestion
	limitVersionBumpToFlag    = "bump-cap"
	limitDependencyBumpToFlag = "dependency-cap"
```
I think we should be consistent with the naming, in this case "cap". If we prefer to call it limit then I think we should change it everywhere (including the code).

src/bumper/bumper.go
```diff
@@ -16,16 +16,41 @@ var ErrNoTags = errors.New("source did not return any tag")
...
 
+// BumperOption is the interface to add options to the bumper.
+//
+//nolint:revive // Following the options pattern here. Renaming it to `option` could lead to a misunderstanding.
+type BumperOption func(bumper *Bumper)
+
 // New creates a new bumper.
-func New(c changelog.Changelog) Bumper {
-	return Bumper{
+func New(c changelog.Changelog, opts ...BumperOption) Bumper {
```
roobre, line 29: I like the options pattern as much as anyone, but in this case, where the only options we can have in the foreseeable future are just plain values... Should we go with simplicity instead and just export the fields?

src/bump/type.go
```diff
@@ -72,3 +81,19 @@ func Bump(version *semver.Version, bt Type) *semver.Version {
 
 	return &v
 }
+
+// NameToType returns the bump type from a string. The string should be from a constant constant of bump.Name
+// or it will return bump.None.
+func NameToType(name string) Type {
```
roobre, line 87: I confess this one is opinionated
```suggestion
func FromName(name string) Type {
```

src/bumper/bumper_test.go
```diff
@@ -16,10 +17,19 @@ type testCase struct {
...
+type testCaseWithCap struct {
+	name          string
+	changelog     changelog.Changelog
+	current       *semver.Version
+	expected      *semver.Version
+	entryCap      bump.Type
+	dependencyCap bump.Type
+}
+
+func TestBumper_Bump_changesOnlyCases(t *testing.T) {
```
roobre, line 29: Is there a reason to have so many, yet so similar test functions? It seems a bit confusing, or tedious to keep those functions similar.

kang-makes: Go linter did not like the length of the functions: https://github.com/newrelic/release-toolkit/pull/128/commits/9fe51a8bd8f95282a614f1f40eabb31e7eca4246#diff-b724875f4a504c8deb98d07b6ff7a128127f4c3d49220ae899e68863f7cf7799R29

So I split them in three. But I can revert that part and add a `nolint` clause.

src/app/nextversion/nextversion.go
kang-makes: Solved in [#2cb3566](https://github.com/newrelic/release-toolkit/pull/128/commits/2cb3566e752425a9d12752ca6a5388b3461d09c4)

src/bumper/bumper.go
kang-makes: As we are not creating a public API I think it does not make sense to have a strong encapsulation. Also improves readability.

Removed in [#8c22e2d](https://github.com/newrelic/release-toolkit/pull/128/commits/8c22e2d1ddfd7549a343c8de3996a1d153d4d202).

src/bump/type.go
kang-makes: I am letting more people review the PR so we can have more opinions on this.

src/bumper/bumper_test.go
kang-makes: I see there are more `nolint` on other tests so I am adding them to these tests too...

src/bump/type.go
sigilioso: If the package name were `type` I would definitely prefer `FromName` because it would be used as `type.FromName`.  As it is in `bumper` package I think having the type in the function name is a bit clearer, another option would be `bumper.TypeFromName`.

src/app/nextversion/nextversion.go
```diff
@@ -100,7 +116,13 @@ func NextVersion(cCtx *cli.Context) error {
 		return err
 	}
 
+	entryCap := bump.NameToType(strings.ToLower(cCtx.String(BumpCapFlag)))
+	dependencyCap := bump.NameToType(strings.ToLower(cCtx.String(DependencyCapFlag)))
```
roobre, line 120: It seems to me here that `NameToType` requiring the string to be lowercase is an implementation detail that could be hidden from the caller. Do you think it makes sense to move `strings.ToLower` inside of `NameToType` instead?

src/bump/type.go
```diff
@@ -72,3 +81,19 @@ func Bump(version *semver.Version, bt Type) *semver.Version {
...
+	//nolint:exhaustive // case NoneName is captured in the last default clause.
```
roobre, line 88: If I call `NameToType("foobar")` right now I'd get `bump.None`, but that's likely not what I want, as I should never been allowed to pass "foobar" to `NameToType` and get away with it. By returning `None`, this function is hiding my mistake. Do you think it would make sense to return an error if the type is not translatable?

src/bumper/bumper.go
```diff
@@ -41,7 +41,7 @@ func (b Bumper) Bump(v *semver.Version) (*semver.Version, error) {
 	for _, d := range b.changelog.Dependencies {
 		dependencyBump = dependencyBump.With(d.BumpType())
 	}
-	dependencyBump = dependencyBump.Cap(b.EntryCap)
+	dependencyBump = dependencyBump.Cap(b.DependencyCap)
```
roobre, line 44: 👍🏻 Good catch!

roobre (commented):
Looking great so far, only a couple of minor comments

src/app/nextversion/nextversion.go
kang-makes: It makes sense. Solved in [#c970483](https://github.com/newrelic/release-toolkit/pull/128/commits/c9704833ba2134da259a84b74334e8dc8dabf717)

src/bump/type.go
kang-makes: If this was for internal application logic as it was until now, I would discuss it... But as it is a parameter that is provided by the user, changing the value silently is a really bad UX. You are completely right.

The change you suggest is solved in [#1453c4a](https://github.com/newrelic/release-toolkit/pull/128/commits/1453c4a6d6cb15f2de7dfa9a2f0b97ef0391055b)

src/app/nextversion/nextversion.go
```diff
@@ -100,7 +115,19 @@ func NextVersion(cCtx *cli.Context) error {
 		return err
 	}
 
+	entryCap, err := bump.NameToType(cCtx.String(BumpCapFlag))
+	if err != nil {
+		return fmt.Errorf("error with the changes' cap: %w", err)
+	}
+	dependencyCap, err := bump.NameToType(cCtx.String(DependencyCapFlag))
+	if err != nil {
+		return fmt.Errorf("error with dependencies' cap: %w", err)
```
roobre, line 124:
```suggestion
		return fmt.Errorf("parsing dependency bump cap: %w", err)
```

src/app/nextversion/nextversion.go
```diff
@@ -100,7 +115,19 @@ func NextVersion(cCtx *cli.Context) error {
 		return err
 	}
 
+	entryCap, err := bump.NameToType(cCtx.String(BumpCapFlag))
+	if err != nil {
+		return fmt.Errorf("error with the changes' cap: %w", err)
```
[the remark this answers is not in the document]
kang-makes: I like more your wording. solved in [#0eb5b56](https://github.com/newrelic/release-toolkit/pull/128/commits/0eb5b567f41c0b40ab9bc50adde5facdb51cae3b)

src/app/nextversion/nextversion.go
kang-makes: I like more your wording. solved in [#0eb5b56](https://github.com/newrelic/release-toolkit/pull/128/commits/0eb5b567f41c0b40ab9bc50adde5facdb51cae3b)

src/bump/type.go
```diff
@@ -72,3 +88,22 @@ func Bump(version *semver.Version, bt Type) *semver.Version {
...
+	case PatchName:
+		return Patch, nil
+	case MinorName:
+		return Minor, nil
+	case MajorName:
+		return Major, nil
+	case NoneName:
+		return None, nil
+	default:
+		return None, ErrNameNotValid
```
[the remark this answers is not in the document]
kang-makes: Solved in [#0eb5b56](https://github.com/newrelic/release-toolkit/pull/128/commits/0eb5b567f41c0b40ab9bc50adde5facdb51cae3b)

Issue #132, opened 2023-01-27
`next-version` should error if there is no previous version 


roobre:
During discussions in #110, we decided that it should be okay for `next-version` to return something, namely `v0.0.1`, if there are no previous versions in the repo. This is useful so a pipeline built on top of the release toolkit can bootstrap a repo from scratch.

However, I think we should revisit this. Let us consider the following case:
- I have a repository with several versions in, and current version is `v1.2.3`
- I make some changes to the pipeline using rt, and I mistakenly add (or change) the `--tag-prefix myproject` argument/input to the `rt` actions
- The release toolkit happily accepts this

On my next release, something bad will happen: `next-version` will see that there are no previous tags matching the prefix, and will return `v0.0.1`. My next release will mistakenly be `v0.0.1`, and I won't realize until the release is done.

I think we should prevent this footgun by default.

### Suggestion

- Make a small change so `next-version` errors by default with a helpful message if no previous versions are found.
- Add a flag to `next-version`, e.g. `--first`, false by default, that will cause `next-version` to return `v0.0.1` if there are no previous versions (current behavior).

---

Eager to hear your thoughts on this :)

sigilioso:
I agree, I would rather prefer to fail. 

However, I am not sure if including a flag to `next-version` to keep the current behavior is a good idea. We could end up in the same situation you described using that flag. Not keeping the current behavior means that pipelines on top of the release toolkit cannot bootstrap a repo without an additional step to create a first or _zero_ version. But, as I see it, that can be tackled in the bootstrap process itself.

alvarocabanas:
I see your point and I agree we should error, but I am with @sigilioso, we should always error and specify in the description of the command and in the documentation that the command always needs a previous tag

roobre:
> Not keeping the current behavior means that pipelines on top of the release toolkit cannot bootstrap a repo without an additional step to create a first or zero version. But, as I see it, that can be tackled in the bootstrap process itself.

Definitely agree with this. Creating a `v0.0.1` tag as part of the repo creation seems acceptable to me.

> However, I am not sure if including a flag to next-version to keep the current behavior is a good idea. We could end up in the same situation you described using that flag.

I think this could be avoided if we make the name very obvious, like `--allow-empty` instead of the more opaque `--first` I proposed.

On the other hand the flag may not be worth having anyway, as like you say there is room for it being misused. I'm okay to drop it if you don't see the value in it.

PR #133, opened 2023-01-27
app/nextversion: make `next-version` error if there are no previous versions

roobre:
As discussed in #132.

Fun fact: I used https://github.com/zurawiki/gptcommit to try to write the commit message for me. Unfortunately that's not what you;'re seeing. The message gptcommit suggested was this ~extremely unrelated piece of crap~:

```
Improve recordings and completions accuracy

- Refactored the authentication code to use a more secure hashing algorithm
- Implemented a new API endpoint for user authentication
- Updated the database schema to store user authentication data securely

```

### DIFF
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -38,8 +38,9 @@ var Cmd = &cli.Command{
 	Usage: "Prints the next version according to the current one, the changelog.yaml file, and semver conventions.",
 	UsageText: `Current version is automatically discovered from git tags in the repository, in semver order. 
 Tags that do not conform to semver standards are ignored.
-Several flags can be specified to limit the set of tags that are scanned, and to override both the current version ` +
-		`being detected and the computed next version.`,
+Several flags can be specified to limit the set of tags that are scanned, and to override both the current version being
+detected and the computed next version.
+next-version will exit with an error if no previous versions are found in the git repository.`,
 	Flags: []cli.Flag{
 		&cli.StringFlag{
 			Name:    tagPrefix,
--- a/src/app/nextversion/nextversion.go
+++ b/src/app/nextversion/nextversion.go
@@ -1,6 +1,7 @@
 package nextversion
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -129,6 +130,13 @@ func NextVersion(cCtx *cli.Context) error {
 	bmpr.DependencyCap = dependencyCap
 
 	next, err := bmpr.BumpSource(versionSrc)
+	if errors.Is(err, bumper.ErrEmptySource) {
+		log.Errorf("Refusing to compute next version as no previous version was found. Please create an initial version first.")
+	}
+
+	if err != nil {
+		return fmt.Errorf("computing next version: %w", err)
+	}
 
 	nextStr := ""
 	switch {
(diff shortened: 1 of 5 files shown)


src/app/nextversion/nextversion.go
```diff
@@ -129,6 +130,13 @@ func NextVersion(cCtx *cli.Context) error {
 	bmpr.DependencyCap = dependencyCap
 
 	next, err := bmpr.BumpSource(versionSrc)
```
roobre, line 132: I think we were previously ignoring this error. Some of the thousand linters should have caught this, right? Or am I missing something?

roobre: I was missing something, it is handled in the switch statement. I added a comment to clarify this.

alvarocabanas (changes requested):
The changes LGTM, but I would add to the cli command description and to the general Readme and the action readme this behavior and/or the need to create a first version tag.

roobre:
Good catch @alvarocabanas! Should be fixed in https://github.com/newrelic/release-toolkit/pull/133/commits/58161659954e7b9381b2fe5601c5b4af66c548b6

src/app/nextversion/nextversion_test.go
```diff
@@ -318,10 +318,48 @@ dependencies:
...
+	}
+}
+
+//nolint:paralleltest // urfave/cli cannot be tested concurrently.
+func TestNextVersion_Fails(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		yaml     string
+		args     string
+		expected string
```
sigilioso, line 358: Nit: we don't need a string expected output in this set of tests at the moment, do we?

src/app/nextversion/nextversion_test.go
```diff
@@ -346,7 +383,7 @@ dependencies:
 		tc := tc
 		//nolint:paralleltest // urfave/cli cannot be tested concurrently.
 		t.Run(tc.name, func(t *testing.T) {
-			repoDir := repoWithTags(t, tc.tags...)
+			repoDir := t.TempDir()
```
sigilioso, line 386: I expected the repository to be initialized with something like

```go
repoDir := repoWithTags(t) // repo created with no tags
```

But now I reckon it to be a different test case.  As I see it, a repository with no tags and a directory (which is not a git repository) are different failure test cases. 

If we wanted to cover both, we could define the repository initialization as one of the test arguments so we could use:

```go
repoDir := tc.repo(t)
```

WDYT?

sigilioso (commented):
Thanks for taking care of it 👍 
I've left a couple of comments related to the tests.

alvarocabanas (approved):
LGTM, once you tackle the comments, for me is good to go

src/app/nextversion/nextversion_test.go
roobre: Good catch!

src/app/nextversion/nextversion_test.go
roobre: I think it's a good point, I should have used a repo with no tags. I'm not sure why I went for the empty folder 🤔 

From what the command is concerned, there shouldn't be any difference though. But... it's easy enough to just provide the function :P

roobre: Done in https://github.com/newrelic/release-toolkit/pull/133/commits/8527223216e8d02eb0c2a17faa5d634e8be0cac3

sigilioso (approved):
🚀

PR #134, opened 2023-01-31
NR-79796: check GitHub links and try with or without leading v

### LINKED ISSUE #72: Github mapper incorrect link when leading v
We have detected that certain dependencies that have releases with leading "v" in the version, when reported by dependabot, the commit message strips the v as following:
- `chore(deps): bump github.com/spf13/viper from 1.7.0 to 1.10.1`

When the Github mapper generates the changelog link, it doesn't work because it should be in the format:
- `https://github.com/spf13/viper/releases/tag/v1.10.1 `

One option we can do is add a step in the mapper that curls the url with and without the leading v and save the one that gets a 200.

sigilioso:
Resolves #72 

A new mapper was introduced to check changelog links and retry the mapping with and without _leading v_. This new mapper wraps the Github mapper to solve #72.

I am not particularly fond of additional libraries to mock http requests (usually `httptest` is enough) but I needed to include [gock](https://github.com/h2non/gock) to mock http requests in the integration tests. An additional dependency is needed but it is only for testing purposes,  it keeps the link integration tests reasonable clean we avoid performing external requests in tests.


.github/workflows/self_test.yaml
```diff
@@ -232,8 +232,8 @@ jobs:
             - name: infrastructure-agent
               to: v1.2.3
             - name: github.com/spf13/viper
-              from: 4.0.3
-              to: 4.1.2
+              from: 1.14.0
+              to: 1.15.0
```
sigilioso, line 236: Using an existing tag works as expected... but I'm not sure if relying on the external tag to perform self-tests is a good idea. Since we are already covering the Github links in the integration tests, maybe it would make sense to remove this dependency from here. WDYT?

src/changelog/linker/mapper/leadingv.go
```diff
@@ -0,0 +1,82 @@
...
+	switchedLiteral := ""
+	if strings.HasPrefix(literal, "v") {
+		switchedLiteral = literal[1:]
+	} else {
+		switchedLiteral = "v" + literal
+	}
+	switchedDep := dep
+	switchedVersion, err := semver.NewVersion(switchedLiteral)
+	if err != nil {
+		log.Debugf("Could not switch leading v to %q valid semver version: %s.", literal, err)
```
roobre, line 61: As this is something that should not be happening, could we promote it to `Errorf` and clarify that is internal?

src/changelog/linker/mapper/leadingv.go
```diff
@@ -0,0 +1,82 @@
...
+// performs a request to the url corresponding to the link and check its status code.
+func WithLeadingVCheck(mapper linker.Mapper) *LeadingVCheck {
+	return &LeadingVCheck{mapper: mapper, checkLink: checkLinkResponse}
+}
+
+func (l *LeadingVCheck) Map(dep changelog.Dependency) string {
+	link := l.mapper.Map(dep)
+	if link == "" || l.checkLink(link) {
+		return link
+	}
```
roobre, line 35: I'd suggest to split this `if` and add some `log.Trace` (or Debug) saying "Skipping link check as inner mapper did not return anything", and "Performing link check on %q".

src/changelog/linker/mapper/leadingv.go
```diff
@@ -0,0 +1,82 @@
...
+
+const (
+	checkTimeoutSeconds = 1
+)
+
+// LeadingVCheck wraps another mapper and performs a check over the result of Map. If the check is not
+// successful, it tries to generate it prepending a leading v to the version.
+type LeadingVCheck struct {
+	mapper    linker.Mapper
+	checkLink func(link string) bool
```
roobre, line 22: I think it is dangerous to reduce the signature of a link check to just a single bit of information (`bool`). Namely, the main problem I see with this is that a network error (e.g. github outage) will be indistinguishable from a tag not existing.

I think we could change this to
```suggestion
	checkLink func(link string) (bool, error)
```

Where errors would be network/transport errors, and modify the code so it logs those very verbosely (Errorf), so users can at least notice that a check is not conclusive.

If the check fails probably we want to return whatever the inner mapper returned.

src/changelog/linker/mapper/leadingv.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return ""
+}
+
+func (l *LeadingVCheck) switchDepLeadingV(dep changelog.Dependency) (changelog.Dependency, bool) {
+	literal := dep.To.Original()
+	switchedLiteral := ""
+	if strings.HasPrefix(literal, "v") {
+		switchedLiteral = literal[1:]
```
roobre, line 54: Slicing strings make me shiver a bit 😅 
```suggestion
		switchedLiteral = strings.TrimPrefix(literal, "v")
```

src/changelog/linker/mapper/leadingv_test.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func TestLeadingVCheck_Map(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name        string
+		wrapped     linker.Mapper
+		dep         changelog.Dependency
+		checkValue  string
```
roobre, line 33: I'm struggling a bit to understand this test. What is `checkValue exactly`?

src/changelog/linker/mapper/leadingv.go
alvarocabanas: I was going to propose to return error in the second param in this function instead of boolean and handle the log or return in the caller.

src/changelog/linker/mapper/leadingv.go
sigilioso: You are right, for some reason I thought that not including the link if the check fails for any reason was OK to skip the changelog link... It might be OK but doing it silently can be very frustrating. Thanks for pointing it out.

Concerning the value to return, I am not sure if I prefer including a link which may be broken (the value that the inner mapper returned) or no link (an empty string). WDYT?

src/changelog/linker/mapper/leadingv.go
sigilioso: Don't be afraid of slices! 😉 (Although, using `TrimPrefix` is way more readable)

src/changelog/linker/mapper/leadingv.go
sigilioso: As it is something that should not happen (if it happens it is because we build a dependency whose `To` field point to an _invalid_ `semver.Version` struct (such as an empty one). I would rather prefer to handle the error here.

src/changelog/linker/mapper/leadingv_test.go
sigilioso: It is the link value the test function expects for each of the test cases. 

It is actually a bit confusing. I started defining a check function for each test case and then I ended up using a _common_ check function and a string instead.

I'll try to make it easier to while tackling the rest of the PR comments.

src/changelog/linker/mapper/leadingv.go
```diff
@@ -0,0 +1,82 @@
...
+type LeadingVCheck struct {
+	mapper    linker.Mapper
+	checkLink func(link string) bool
+}
+
+// WithLeadingVCheck returns a LeadingVCheck with the provided underlying mapper and a check function which
+// performs a request to the url corresponding to the link and check its status code.
+func WithLeadingVCheck(mapper linker.Mapper) *LeadingVCheck {
+	return &LeadingVCheck{mapper: mapper, checkLink: checkLinkResponse}
+}
```
kang-makes, line 29: NIT
```suggestion
// NewWithLeadingVCheck returns a LeadingVCheck with the provided underlying mapper and a check function which
// performs a request to the url corresponding to the link and check its status code.
func NewWithLeadingVCheck(mapper linker.Mapper) *LeadingVCheck {
	return &LeadingVCheck{mapper: mapper, checkLink: checkLinkResponse}
}
```

README.md
```diff
@@ -201,6 +201,15 @@ dependencies:
       changelog: https://github.com/my-org/a-dependency/releases/tag/3.0.0
 ```
 
+#### Automatically detected GitHub repository
+
+When the dependency does not match any entry in the dictionary file but it does match GitHub repositories (`github.com/<org><repo>`),
```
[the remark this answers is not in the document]
sigilioso: Thank you!

.github/workflows/self_test.yaml
sigilioso: Thanks, done here: [dc67e07](https://github.com/newrelic/release-toolkit/pull/134/commits/dc67e07f5c7412344599d6161c6708350dfc70a5)

src/changelog/linker/mapper/leadingv.go
sigilioso: I included some logs here: [6f5cbde](https://github.com/newrelic/release-toolkit/pull/134/commits/6f5cbde1ffe7074c0185bed72543d9d8264ec1e3)

src/changelog/linker/mapper/leadingv.go
sigilioso: Whatever we decide to do when links cannot be checked due to unexpected errors, that situation should be handled by the mapper and not by the check function, so I changed the function signature as you suggested:[6f5cbde](https://github.com/newrelic/release-toolkit/pull/134/commits/6f5cbde1ffe7074c0185bed72543d9d8264ec1e3)

Regarding the potentially broken links, there is a risk of generating those when the check cannot be performed due to network issues, but as long as we properly warn the users it could be a good a approach.

src/changelog/linker/mapper/leadingv_test.go
sigilioso: I changed it back. Now the `checkLink` function is directly defined and I think the test cases are a bit easier to follow. Changed here: [dc67e07](https://github.com/newrelic/release-toolkit/pull/134/commits/dc67e07f5c7412344599d6161c6708350dfc70a5). Let me know your thoughts.

src/changelog/linker/mapper/leadingv.go
sigilioso: Changed here: [6f5cbde](https://github.com/newrelic/release-toolkit/pull/134/commits/6f5cbde1ffe7074c0185bed72543d9d8264ec1e3)

src/changelog/linker/mapper/leadingv.go
roobre: I think there is no good solution for this one. If we remove the link (return empty) on error, we could be reflecting transient errors on a permanent state (the changelog). If we error out and abort, we would be making the tool unreliable. If we do nothing (return link as-is from the inner mapper), we could be creating an inconsistency.

I don't like either, but I think returning as-is and logging an error is the one I dislike less.

src/app/link/link.go
```diff
@@ -87,7 +87,8 @@ func Link(cCtx *cli.Context) error {
 		mappers = append(mappers, dic)
 	}
 
-	mappers = append(mappers, mapper.Github{})
+	githubMapper := mapper.WithLeadingVCheck(mapper.Github{})
+	mappers = append(mappers, githubMapper)
```
roobre, line 91: Given that this feature requires network access, I think we should at least make it opt-out so it can run on restricted environments.

src/changelog/linker/mapper/leadingv.go
kang-makes: As far as I know, it is idiomatic to use `New` when you get a struct ready to be used in the same way constructors exist in OOP languages.

To have more than one constructor in Go we should use `NewWithWhatever`. Mapper here only has one constructor but if you want to be explicit I would prefer the `NewWithLeadingVCheck` in this case.

I am used to name functions that start with `With` for options using the options pattern: https://golang.cafe/blog/golang-functional-options-pattern.html

sigilioso: Fair enough! 😃

src/app/link/link.go
sigilioso: New option included here: [8161372](https://github.com/newrelic/release-toolkit/pull/134/commits/81613724263fb3cc885d3ad6e1594b11a47c6cee)

src/changelog/linker/mapper/leadingv.go
sigilioso: Updated here: [1315f5a](https://github.com/newrelic/release-toolkit/pull/134/commits/1315f5a7539fd21d8b4795257c35bc609d2bd1d5)

kang-makes (approved):

alvarocabanas (approved):
LGTM

src/app/link/link.go
```diff
@@ -41,6 +42,14 @@ var Cmd = &cli.Command{
 			EnvVars: common.EnvFor(sampleFlag),
 			Usage:   "Prints a sample dictionary to stdout.",
 		},
+		&cli.BoolFlag{
+			Name:    disableGithubValidationFlag,
+			EnvVars: common.EnvFor(disableGithubValidationFlag),
+			Usage: "Disables changelog links validation for automatically detected Github repositories. " +
+				"When disabled, changelog links for Github repositories are directly rendered using " +
+				"https://github.com/<org>/<repo>/releases/tag/<new-version> with no validation, so no external request are performed.",
```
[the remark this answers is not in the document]
sigilioso: It definitely makes sense, I updated the description here: [6469731](https://github.com/newrelic/release-toolkit/pull/134/commits/646973186119fc30c5f11c31419757768464322c)

PR #135, opened 2023-01-31
Add contrib action for OHI release (NR-83896)

kang-makes:
There is a workflow that is common for all the OHIs that it would be useful to have standardized in a single action.

This contribution enables an action and a script to run locally that includes all the actions that are common to the release of an OHI.

~~This PR is in draft as I must have the code uploaded to be able to test it in antoher repository.~~ Test was a success: newrelic/nri-kafka#207

Screenshot 2023-02-01 at 12 48 08

### Notes

To test the script of this branch, instead of the command from the readme that pointes to the ~~`main`~~ `v1` tag, you can test with `curl -L https://raw.githubusercontent.com/newrelic/release-toolkit/feat/contrib/ohi-release/contrib/ohi-release/run.sh | bash`

### DIFF
--- a/.github/workflows/self_test.yaml
+++ b/.github/workflows/self_test.yaml
@@ -518,15 +518,19 @@ jobs:
         with:
           git-root: ${{ env.MOCK_REPO_ACTION }}
       - name: Run script
-        working-directory: ${{ env.MOCK_REPO_SCRIPT }}
-        run: $(git rev-parse --show-toplevel)/contrib/ohi-release/run.sh
+        run: |
+          cd "${{ env.MOCK_REPO_SCRIPT }}"
+          # CLI autodetects whether it is running on GHA or not. If we let GHA set this to true, as it normally does,
+          # tests cases where GITHUB_ACTIONS is expected to be unset would fail.
+          GITHUB_ACTIONS=false "${GITHUB_WORKSPACE}/contrib/ohi-release/run.sh"
       - name: Action outputs
         working-directory: ${{ env.ACTION_OUTPUTS }}
         run: |
           echo "${{ steps.ohi-release.outputs.changelog }}" > CHANGELOG.md
           echo "${{ steps.ohi-release.outputs.changelog-partial }}" > CHANGELOG.partial.md
 
-      - run: |
+      - name: Asserts
+        run: |
           exit_status=0
 
           if ! cmp --silent expected-changelog.md ${{ env.MOCK_REPO_ACTION }}/CHANGELOG.md; then
--- a/.github/workflows/self_test.yaml
+++ b/.github/workflows/self_test.yaml
@@ -428,8 +428,8 @@ jobs:
             exit 1
           fi
 
-  contrib-ohi-release:
-    name: validate ohi-release (contrib action)
+  contrib-ohi-release-notes:
+    name: validate ohi-release-notes (contrib action)
     runs-on: ubuntu-latest
     env:
       MOCK_REPO_ACTION: ./mock_repo_action
@@ -513,8 +513,8 @@ jobs:
           EOF
 
       - name: Run action
-        uses: ./contrib/ohi-release
-        id: ohi-release
+        uses: ./contrib/ohi-release-notes
+        id: ohi-release-notes
         with:
           git-root: ${{ env.MOCK_REPO_ACTION }}
       - name: Run script
@@ -522,12 +522,12 @@ jobs:
           cd "${{ env.MOCK_REPO_SCRIPT }}"
           # CLI autodetects whether it is running on GHA or not. If we let GHA set this to true, as it normally does,
           # tests cases where GITHUB_ACTIONS is expected to be unset would fail.
-          GITHUB_ACTIONS=false "${GITHUB_WORKSPACE}/contrib/ohi-release/run.sh"
+          GITHUB_ACTIONS=false "${GITHUB_WORKSPACE}/contrib/ohi-release-notes/run.sh"
       - name: Action outputs
         working-directory: ${{ env.ACTION_OUTPUTS }}
         run: |
-          echo -ne "${{ steps.ohi-release.outputs.changelog }}" > CHANGELOG.md
-          echo -ne "${{ steps.ohi-release.outputs.changelog-partial }}" > CHANGELOG.partial.md
+          echo -ne "${{ steps.ohi-release-notes.outputs.changelog }}" > CHANGELOG.md
+          echo -ne "${{ steps.ohi-release-notes.outputs.changelog-partial }}" > CHANGELOG.partial.md
 
       - name: Asserts
         run: |
--- a/.github/workflows/self_test.yaml
+++ b/.github/workflows/self_test.yaml
@@ -427,3 +427,138 @@ jobs:
             echo "valid should have returned false" >&2
             exit 1
           fi
+
+  contrib-ohi-release:
+    name: validate ohi-release (contrib action)
+    runs-on: ubuntu-latest
+    env:
+      MOCK_REPO_ACTION: ./mock_repo_action
+      MOCK_REPO_SCRIPT: ./mock_repo_script
+      ACTION_OUTPUTS: ./action_outputs
+    steps:
+      - uses: actions/checkout@v3
+      - name: Configure test repo
+        shell: bash
+        run: |
+          # MOCK_REPO_SCRIPT is not needed as it is copied from ACTION one.
+          mkdir -pv "${MOCK_REPO_ACTION}" "${ACTION_OUTPUTS}"
+          cd "${MOCK_REPO_ACTION}"
+
+          git config --global user.email "you@example.com"
+          git config --global user.name "Your Name"
+
+          git init
+          cat > CHANGELOG.md <<EOF
+          # Changelog
+            This is based on blah blah blah
+
+          ## v1.2.3 - 20YY-DD-MM
+
+          ### Enhancements
+          - This is in the past and should be preserved
+          EOF
+
+          git add CHANGELOG.md
+          for i in 1 2 3; do
+            touch file$i
+            git add file$i
+            git commit -m commit_$i
+          done
+          git tag 1.2.3
+
+
+          cat > CHANGELOG.md <<EOF
+          # Changelog
+            This is based on blah blah blah
+
+          ## Unreleased
+          ### Enhancement
+          - test
+
+          ## v1.2.3 - 20YY-DD-MM
+
+          ### Enhancements
+          - This is in the past and should be preserved
+          EOF
+
+          cd $OLDPWD
+          cp -r "${MOCK_REPO_ACTION}" "${MOCK_REPO_SCRIPT}"
+      - name: Expected CHANGELOG
+        shell: bash
+        run: |
+          cat > expected-changelog.md <<EOF
+          # Changelog
+            This is based on blah blah blah
+
+          ## Unreleased
+
+          ## v1.3.0 - $(date +%Y-%m-%d)
+
+          ### 🚀 Enhancements
+          - test
+
+          ## v1.2.3 - 20YY-DD-MM
+
+          ### Enhancements
+          - This is in the past and should be preserved
+          EOF
+      - name: Expected partial CHANGELOG
+        shell: bash
+        run: |
+          cat > expected-partial.md <<EOF
+          ## v1.3.0 - $(date +%Y-%m-%d)
+
+          ### 🚀 Enhancements
+          - test
+          EOF
+
+      - name: Run action
+        uses: ./contrib/ohi-release
+        id: ohi-release
+        with:
+          git-root: ${{ env.MOCK_REPO_ACTION }}
+      - name: Run script
+        working-directory: ${{ env.MOCK_REPO_SCRIPT }}
+        run: $(git rev-parse --show-toplevel)/contrib/ohi-release/run.sh
+      - name: Action outputs
+        working-directory: ${{ env.ACTION_OUTPUTS }}
+        run: |
+          echo "${{ steps.ohi-release.outputs.changelog }}" > CHANGELOG.md
+          echo "${{ steps.ohi-release.outputs.changelog-partial }}" > CHANGELOG.partial.md
+
+      - run: |
+          exit_status=0
+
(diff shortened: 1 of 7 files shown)


contrib/ohi-release/action.yml
```diff
@@ -0,0 +1,74 @@
...
+    required: true
+    default: '.github'
+  fail-if-held:
+    description: 'Fail if the held toggle is active'
+    required: true
+    default: 'true'
+  link-dependencies-enable:
+    description: 'Enables/disables link dependencies step'
+    required: true
+    default: 'false'
```
kang-makes, line 15: I had to add a toggle to the `link-dependencies` because it fails if the file `.github/rt-dictionary.yaml` does not exist.

I **think** that it is more like in OHIs to not link to any dependency that linking them... That is the reason I default to `false` but we can discuss this during the review.

contrib/ohi-release/README.md
```diff
@@ -0,0 +1,63 @@
...
+```yaml
+- uses: newrelic/release-toolkit/contrib/ohi-release@v1
+  id: release
+- name: Commit updated changelog
+  run: |
+    git add CHANGELOG.md
+    git commit -m "Update changelog with changes from ${{ steps.release.outputs.next-version }}"
+    git push -u origin main
+    gh release create ${{ steps.release.outputs.release-title }} --target $(git rev-parse HEAD) --notes-file CHANGELOG.partial.md
+```
```
alvarocabanas, line 16: Nice! it really simplifies everything for us 👍

contrib/ohi-release/action.yml
```diff
@@ -0,0 +1,74 @@
...
+  release-changelog:
+    description: "Complete changelog of this release"
+    value: ${{ steps.release.outputs.changelog }}
+  release-changelog-partial:
+    description: "Changelog for only this release"
+    value: ${{ steps.release.outputs.changelog-partial }}
+runs:
+  using: composite
+  steps:
+    - uses: newrelic/release-toolkit/validate-markdown@v1
```
alvarocabanas, line 36: This is not required here, the validate-markdown is supposed to be executed as a separate action on the PR workflow or individually executing rt validate-markdown in your local environment.
We might add it here as an extra validation if you decide is valuable, because it might be useful to have it if executing the shell locally.

contrib/ohi-release/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Parameters
+
+All parameters are optional:
+  * `excluded-dirs` exclude commits whose changes only impact files in specified dirs relative to repository root. Defaults to ".github".
```
alvarocabanas, line 21: Why not adding also the included-dirs flag? At the moment we don't have any case where we want to include only one folder instead of excluding tones of them, but it could happen.

contrib/ohi-release/action.yml
```diff
@@ -0,0 +1,74 @@
+name: OHI release
+description: Wrapper for release toolkit that runs commands needed to release an OHI
+inputs:
+  excluded-dirs:
+    description: 'Exclude commits whose changes only impact files in specified dirs relative to repository root. Defaults to ".github".'
+    required: true
```
alvarocabanas, line 6: I would leave all this parameters as required false.

contrib/ohi-release/action.yml
alvarocabanas: The thing is, we might think if we want the link-dependencies-dictionary to only throw a warning if the file is not found because this command also adds the changelog automatically for git dependencies, and I would like to always have that that feature

alvarocabanas (commented):
Good work! I left some comments

contrib/ohi-release/action.yml
kang-makes: I found issues creating the markdown way below because the markdown was not valid and this allows the user to not footgun itself.

Here is a test case:
Given a Changelog (snippet) like this
```markdown
## Unreleased
### Changes
test
```
If I try to `validate-markdown` it returns `unreleased changelog can't only contain notes`

But if I run `generate-yaml` it returns a valid YAML:
```yaml
notes: |-
    ### Changes
    test
changes: []
dependencies: []
```

There are lots of tests that I am able to pass, like using this markdown:
```markdown
## Unreleased
### Changes
test
### Enhancement
- A test in singular
### Enhancements
- A test in plural
```

But is a safety measurement until we add better testing when we footgun ourselves a couple of times 🤣

contrib/ohi-release/README.md
kang-makes: That is the point I did not add it. I did not want to expose all the flags possible for all the commands that this wrapper runs until it is needed.

It you can barely thing a single use case I'll be happy to add it :)

contrib/ohi-release/action.yml
kang-makes: After taking some time to think about it, from my point of view, the CLI UX is correct: If you set a flag with a path to a file it should fail if it does not exist or you cannot parse it.

As the approach I chose was not correct, I changed the default in [#b75bd12](https://github.com/newrelic/release-toolkit/pull/135/commits/b75bd12b5be2dc02d4f5aae737ed14a9efc90c6f)

contrib/ohi-release/action.yml
kang-makes: Solved in [#b75bd12](https://github.com/newrelic/release-toolkit/pull/135/commits/b75bd12b5be2dc02d4f5aae737ed14a9efc90c6f)

contrib/ohi-release/README.md
```diff
@@ -0,0 +1,63 @@
...
+  * `fail-if-held` fails if the held toggle is active
```
sigilioso, line 22:
```suggestion
  * `fail-if-held` fails if the held toggle is active. Defaults to `true`.
```

contrib/ohi-release/README.md
```diff
@@ -0,0 +1,63 @@
+# 🛠️ `ohi-release`
+
+This is a wrapper of all the steps needed to update the changelog and render a snippet to be ready to be used as a release message. This contribution also includes a script that allow to replicate what this action does locally.
+
+## Example Usage
+
+```yaml
+- uses: newrelic/release-toolkit/contrib/ohi-release@v1
```
sigilioso, line 8: As the release is supposed to be created in a different step (using the action output) would it make sense to name the action `contrib/ohi-release-notes` instead of `contrib/ohi-release`?

sigilioso (commented):
🚀 This will help us a lot, thanks!

I'm not approving it yet because it is still a draft PR. I've left two small suggestions.

contrib/ohi-release/README.md
kang-makes: Fair enough. Fixed in [#2e447e6](https://github.com/newrelic/release-toolkit/pull/135/commits/2e447e6f41457d446032646f59654b6ce83c963f)

contrib/ohi-release/README.md
kang-makes: Fixed in [#2e447e6](https://github.com/newrelic/release-toolkit/pull/135/commits/2e447e6f41457d446032646f59654b6ce83c963f)

kang-makes:
Blocked by #139

kang-makes:
Rebased :)

contrib/ohi-release-notes/action.yml
```diff
@@ -0,0 +1,94 @@
...
+    default: "."
+  excluded-dirs:
+    description: Exclude commits whose changes only impact files in specified dirs relative to repository root. Defaults to ".github".
+    default: '.github'
+  fail-if-held:
+    description: Fail if the held toggle is active. Defaults to `true`.
+    default: 'true'
+  link-dependencies-dictionary:
+    description: Sets the link dependency dictionary. Defaults to ".github/rt-dictionary.yaml".
+    default: ''
```
sigilioso, line 15: Maybe there is some magic there and I'm wrong, but I think the

> Defaults to ".github/rt-dictionary.yaml"

and the corresponding readme comment

don't match with the `default: ''`

sigilioso (approved):
Thanks again! 🚀 

LGTM, simply left a couple of nits.

contrib/ohi-release-notes/run.sh
```diff
@@ -0,0 +1,118 @@
...
+TEMP_BIN=$(mktemp -dt release-toolkit-XXX)
+function cleanup() {
+    rm -rf $TEMP_BIN    || true
+    rm "${GIT_ROOT}/CHANGELOG.md.bak" || true
+    rm "${GIT_ROOT}/changelog.yaml"   || true
+}
+trap cleanup EXIT
+
+
+# usage and helm command. It is also an error exit in case flags are not correct.
```
paologallinaharbur, line 19: helm -> help

contrib/ohi-release-notes/run.sh
```diff
@@ -0,0 +1,118 @@
...
+            --no-fail ) IS_HELD_FAIL=""; shift ;;
+            # Flags for link-dependencies
+            --dictionary ) DICTIONARY="$2"; shift 2 ;;
+            * ) help "option is not accepted/parsable: \"$1\"" ;;
+        esac
+    fi
+done
+
+# building rt
+GOBIN="${TEMP_BIN}" go install ${RT_PKG}
```
paologallinaharbur, line 91: if I am not mistaken when you run an action you download the whole folder, should it try to run the same code? Otherwise we could end up running @latest instead of @v1

.github/workflows/self_test.yaml
```diff
@@ -427,3 +427,142 @@ jobs:
...
+          # CLI autodetects whether it is running on GHA or not. If we let GHA set this to true, as it normally does,
+          # tests cases where GITHUB_ACTIONS is expected to be unset would fail.
+          GITHUB_ACTIONS=false "${GITHUB_WORKSPACE}/contrib/ohi-release-notes/run.sh"
+      - name: Action outputs
+        working-directory: ${{ env.ACTION_OUTPUTS }}
+        run: |
+          echo -ne "${{ steps.ohi-release-notes.outputs.changelog }}" > CHANGELOG.md
+          echo -ne "${{ steps.ohi-release-notes.outputs.changelog-partial }}" > CHANGELOG.partial.md
+
+      - name: Asserts
```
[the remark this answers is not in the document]
kang-makes: Sure! Done here: [#6f8e3b9](https://github.com/newrelic/release-toolkit/pull/135/commits/6f8e3b90c18e065778d5985240164ea751e9c2fc)

contrib/ohi-release-notes/action.yml
kang-makes: It used to be that default but I had to change it and forgot to remove it from the description.

Fixed in [#345ad7a](https://github.com/newrelic/release-toolkit/pull/135/commits/345ad7a05cd21e2d2aaf0f3f46b423161917de2e)

contrib/ohi-release-notes/run.sh
kang-makes: I forgot how to English and my fingers write things without thinking... Fixed in [#345ad7a](https://github.com/newrelic/release-toolkit/pull/135/commits/345ad7a05cd21e2d2aaf0f3f46b423161917de2e)

Good catch!

contrib/ohi-release-notes/run.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+
+set -euo pipefail
+
+RT_PKG="github.com/newrelic/release-toolkit@latest"
```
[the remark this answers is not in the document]
kang-makes: No. The action and the script are two different things.

The action runs the `action.yml` that runs other actions so it stays with the same version that the action has as it is already downloaded. Here you have the line I am referring to: https://github.com/newrelic/release-toolkit/pull/135/files#diff-3435321f89580c19c1ccf925d4992905d93df0e2f863bff8b1ad6261e0cdb7d2R29-R35

note that the `uses` is local because it is already downloaded.

The script is meant to be used locally on our laptops for manual usage. I have to kid you and say RTFM: https://github.com/newrelic/release-toolkit/pull/135/files#diff-286a0a98212926ebee9d0bcce9e6bba72c6987271fbde8b5a5a94c2d9098ad2fR34-R40

contrib/ohi-release-notes/run.sh
kang-makes: This is [completely related to the previous message you left](https://github.com/newrelic/release-toolkit/pull/135#discussion_r1104163193) so I am closing this conversation.

contrib/ohi-release-notes/run.sh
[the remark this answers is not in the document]
paologallinaharbur: Oh, I reviewed it quickly and I missed that, then why having twice the same logic? Can't the action call the script?

I am a bit scared by possible discrepancies: i.e. "There is a run.sh script that **should do** the same as this action"

However, if you discussed it already multiple times and that was the best option I am fine with it, I just find it a bit weird and a possible source of bugs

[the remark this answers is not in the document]
kang-makes: It is inspired by the infrastructure-bundle: https://github.com/newrelic/infrastructure-bundle/blob/4864fbd/run-ci-locally.sh

GitHub gives us a CI but not a way to run it locally. The project `act` (https://github.com/nektos/act) has a good project and it is growing really fast but it is still not a viable solution. So we have a YAML to run an action and another script to run it locally.

There is nothing more we can do. The script also works different from the action as the action needs outputs while the script simply leaves the files to be used and prints the title so the user can _do things_ with that.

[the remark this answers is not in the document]
paologallinaharbur: yes, I agree that running the action locally is not desirable. 

What I meant was having the action running the same script that we can use locally, something like:

```
inputs:
[...]
runs:
  using: "composite"
  steps:
    - run: ./run.sh
      shell: bash
```

So that we are sure that the logic is the same. The inputs could still not be mapped correctly, but at least everything else would be the same and we could remove a whole file. 

However, as I said before I do not want to be a blocker for this in case you have discussed for a while the whole behavior  😄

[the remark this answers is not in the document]
kang-makes: Solved with an offline conversation.

> There are so many differences between [running outside and inside the CI] that it is better to have directly two codebases.

PR #136, opened 2023-02-02
dependabot,renovate: handle revert commits

roobre:
This use case was revealed to me in a dream.

Well, actually not, it was revealed to me yesterday at 23:50 when I had to revert a renovate PR that broke the ingress controller of both of my cluster. But that doesn't sound nearly as deep.

Code is not particularly pretty, mainly because revert commits may be authored by anyone, so we have to ignore the author check. I think this is probably fine: Odds that a commit starts with `Revert` but also happens to match the dependabot/renovate regexes seems quite slim.

I unfortunately had to add a `nolint` directive for the cyclomatic complexity of the `Changelog()` funcitons. I honestly care about cognitive complexity a lot, but not so much for cyclo. LMK if you agree with this, if you don't maybe we can think a way to refactor it.

While developing this, I also figured out that unit test for Renovate is buggy. I'll open a different PR for that.


alvarocabanas (approved):
LGTM, as you said is not so pretty, but it's needed and how you did it is clear, better to not overcomplicate the code.

PR #139, opened 2023-02-07
Add new line at the end of the renderer (NR-83890)

kang-makes:
PR #135 fails because the renderer renders release notes snippets without a line jump at the end.

It is already documented here as a known issue: https://github.com/newrelic/release-toolkit/blob/55454a9e7e4e3a4a80e5ce03ed0563e5fbfaafc7/src/changelog/renderer/renderer.go#L52-L55

But we should add a line jump at the end of the render command in the same way that the merger adds here: https://github.com/newrelic/release-toolkit/blob/55454a9e7e/src/changelog/sources/markdown/merger/merger.go#L66

### DIFF
--- a/src/app/render/render.go
+++ b/src/app/render/render.go
@@ -94,5 +94,14 @@ func Render(cCtx *cli.Context) error {
 		return fmt.Errorf("rendering changelog: %w", err)
 	}
 
+	// At the time of writing this, this is a comment you can read in the changelog renderer:
+	// > For templates to be sane and readable, we need to put spaces between sections _after_ them. This comes with the
+	// > problem of the last section of the doc also printing those spaces, leading to two empty newlines.
+	// Because of that, rendered will output the markdown without a leading line jump that we have to add here.
+	_, err = fmt.Fprintf(mdFile, "\n")
+	if err != nil {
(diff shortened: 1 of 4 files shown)


src/app/render/render.go
```diff
@@ -94,5 +94,14 @@ func Render(cCtx *cli.Context) error {
 		return fmt.Errorf("rendering changelog: %w", err)
 	}
 
+	// At the time of writing this, this is a comment you can read in the changelog renderer:
+	// > For templates to be sane and readable, we need to put spaces between sections _after_ them. This comes with the
+	// > problem of the last section of the doc also printing those spaces, leading to two empty newlines.
+	// Because of that, rendered will output the markdown without a leading line jump that we have to add here.
+	_, err = fmt.Fprintf(mdFile, "\n")
```
alvarocabanas, line 101: I would formulate this comment different because it was hard to follow for me, something like:

_As it's written in a comment in the renderer.Render function, we are putting spaces between sections \_after\_ them, this comes with the problem of the last section of the doc also printing those spaces, leading to two empty newlines, because of that renderer.Render will output the markdown without a leading line jump, so we add it here._

alvarocabanas (approved):
LGTM, I just left a nit

PR #141, opened 2023-03-03
fix(contrib): using static reference

paologallinaharbur:
We cannot use local actions, otherwise, the actions fail saying:
> Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/nri-kafka/nri-kafka/validate-markdown'. Did you forget to run actions/checkout before running your local action?

I do not like that much to point to a specific tag, but I was not able to find a better solution not checking out the action code

### DIFF
--- a/contrib/ohi-release-notes/action.yml
+++ b/contrib/ohi-release-notes/action.yml
@@ -30,11 +30,11 @@ runs:
   using: composite
   steps:
     - name: Validate that the markdown is correct
-      uses: ./validate-markdown
+      uses: newrelic/release-toolkit/validate-markdown@v1
       with:
         markdown: ${{ inputs.git-root }}/CHANGELOG.md
     - name: Generate YAML
-      uses: ./generate-yaml
+      uses: newrelic/release-toolkit/generate-yaml@v1
       with:
         excluded-dirs: ${{ inputs.excluded-dirs }}
         git-root: ${{ inputs.git-root }}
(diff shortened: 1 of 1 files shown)


contrib/ohi-release-notes/action.yml
```diff
@@ -30,19 +30,19 @@ runs:
   using: composite
   steps:
     - name: Validate that the markdown is correct
-      uses: ./validate-markdown
+      uses: newrelic/release-toolkit/validate-markdown@v1
```
paologallinaharbur, line 33: I am open to different solutions if available, I did a "quick fix" to unblock myself on the kafka PR

kang-makes: Oh! Good point, I did not think about that!

There are environment variables that are available only inside an action and there is one that is handy in composite actions: `GITHUB_ACTION_PATH` ([reference](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables))

Can you test this?

```suggestion
      uses: ${{ env.GITHUB_ACTION_PATH }}/validate-markdown
```

paologallinaharbur: Hello! I tested it getting the error:
```
Unrecognized named-value: 'env'. Located at position 1 within expression: env.GITHUB_ACTION_PATH
```

I believe that the `env` is not available in the `uses` context 
 - https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

Docs are not super clear about that though (`uses` keyword is not specified and neither composite actions  )

paologallinaharbur: I am currently checking if we can make something static similar to the GITHUB_ACTION_PATH value.

Otherwise the only way I see to avoid having @v1 is to move this to a script 😖

paologallinaharbur: After talking offline with @kang-makes we decided to keep it like this to avoid to implement a new script

kang-makes (approved):
LGTM.

Just create the [issue](https://github.com/newrelic/release-toolkit/pull/141/files#r1126505914) so we follow it.

PR #143, opened 2023-03-07
feat(newAction): is-empty

paologallinaharbur:
This new command allows to run is-empty to check if the changelog is empty and therefore no automated release should be triggered

### DIFF
--- a/.github/workflows/self_test.yaml
+++ b/.github/workflows/self_test.yaml
@@ -50,7 +50,74 @@ jobs:
           yaml: ${{ env.MOCK_REPO }}/changelog.yaml
       - run: |
           if [[ "${{ steps.held.outputs.is-held }}" != "true" ]]; then
-            echo "is-held should have returned true" >&2
+            echo "is-held should have returned false" >&2
+            exit 1
+          fi
+
+  is-empty:
+    name: is-empty action
+    runs-on: ubuntu-latest
+    env:
+      MOCK_REPO: ./mock_repo
+    steps:
+      - uses: actions/checkout@v3
+      - name: Configure test repo
+        id: repo
+        shell: bash
+        run: |
+          mkdir $MOCK_REPO
+          cd $MOCK_REPO
+
+          git config --global user.email "you@example.com"
+          git config --global user.name "Your Name"
+
+          git init
+          for i in 1 2 3; do
+            touch file$i
+            git add file$i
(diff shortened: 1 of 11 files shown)


.github/workflows/self_test.yaml
```diff
@@ -50,7 +50,74 @@ jobs:
           yaml: ${{ env.MOCK_REPO }}/changelog.yaml
       - run: |
           if [[ "${{ steps.held.outputs.is-held }}" != "true" ]]; then
-            echo "is-held should have returned true" >&2
+            echo "is-held should have returned false" >&2
```
paologallinaharbur, line 53: false-> true

src/app/isempty/isempty.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return fmt.Errorf("loading changelog from file: %w", err)
+	}
+
+	_, _ = fmt.Fprintf(cCtx.App.Writer, "%v\n", ch.Empty())
+
+	gh.SetOutput(isEmptyOutput, ch.Empty())
+	if cCtx.Bool("fail") && ch.Empty() {
+		return cli.Exit("changelog is empty", 1)
+	}
```
sigilioso, line 55: Nit: since we are using the function several times we may store its value.

```suggestion
	chEmpty := ch.Empty()
	_, _ = fmt.Fprintf(cCtx.App.Writer, "%v\n", chEmpty)

	gh.SetOutput(isEmptyOutput, chEmpty)
	if cCtx.Bool("fail") && chEmpty {
		return cli.Exit("changelog is empty", 1)
	}
```

src/app/isempty/isempty.go
```diff
@@ -0,0 +1,58 @@
...
+const isEmptyOutput = "is-empty"
+
+const failFlag = "fail"
+
+// Cmd is the cli.Command object for the is-empty command.
+//
+//nolint:gochecknoglobals
+var Cmd = &cli.Command{
+	Name:  "is-empty",
+	Usage: "Outputs whether automated releases is not needed since changelog is empty",
```
sigilioso, line 23:
```suggestion
	Usage: "Outputs whether automated releases are not needed since changelog is empty",
```

sigilioso (approved):
LGTM, I left a couple of _nits_

PR #146, opened 2023-03-12
app/update-markdown: reflect in cli.Flag that --version is not optional

roobre:
Previous description and `Required` attributed hinted that the `--version` argument to the `update-markdown` command is optional, which is not the case.


sigilioso (commented):
We need to update the [README docs](https://github.com/newrelic/release-toolkit/blob/77b46db031b9e3ea35179cb64f5d17d011faf60d/README_CLI.md?plain=1#L95) as well, don't we?

PR #150, opened 2023-04-18
fix(ohi-release-notes): it should not fail if changelog is empty

paologallinaharbur:
Generate YAML should not fail if the changelog is empty

>   exit-code:
    description: Exit code if changelog is empty
    required: false
    default: “1”


marcsanmi (approved):
Thanks for addressing it 👍

PR #152, opened 2023-05-04
Add dictionary (NR-112268)

gsanchezgavier:
- Adds default dictionary file to be used by the action and the script.
- It removes the dictionary input from the action input to avoid complexity (we think that all integrations will share the same file), still it can be tested with another dictionary by the `--dictionary` flag from the script.

Note that tests are passiging because of [6066c50](https://github.com/newrelic/release-toolkit/pull/152/commits/6066c508cc0369b4b8271fb80a2eccb28bf80b1f)

Following this PR:
- remove dictionary from infra-checkers
- remove infra checkers step from reusable workflow


contrib/ohi-release-notes/run.sh
```diff
@@ -95,6 +98,11 @@ if ! [ -x $RT_BIN ]; then
     help "rt binary is not executable: \"${RT_BIN}\""
 fi
 
+# fetch default dictionary by default
+if [ -z "$DICTIONARY" ]; then
+    DICTIONARY="${TEMP_BIN}/rt-dictionary.yml"
```
kang-makes, line 103: It was called `TEMP_BIN` because it was meants to be used only for the bin.

if it is going to become a temporary folder I would rename it to `TEMP_DIR`.
```suggestion
    DICTIONARY="${TEMP_DIR}/rt-dictionary.yml"
```

kang-makes (approved):
A couple of NITs but LGTM.

Thanks!

contrib/ohi-release-notes/rt-dictionary.yml
kang-makes, line 1: I am comparing it with the list that we have in coreint automation and it seems that there are a couple of links missing. I don't know if this is intended:
 * `nri-docker` might be needed for ECS releases
 * infra bundle might be needed for ECS releases
 * Oracle, nri-prometheus (even if it is deprecated we can release something related to vulnerabilities), varnish, winservices and the Prometheus manager are missing.

contrib/ohi-release-notes/run.sh
```diff
@@ -95,6 +98,11 @@ if ! [ -x $RT_BIN ]; then
...
+    curl -s -o $DICTIONARY $DICTIONARY_URL
```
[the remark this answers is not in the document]
gsanchezgavier: had to force push to sign commits , but is addressed thanks!

contrib/ohi-release-notes/rt-dictionary.yml
gsanchezgavier: i didn't even take a look at this file, thanks.
had to force push to sign commits , but is addressed thanks!

kang-makes (dismissed):
Because of the changes, force-pushes and such I have not reviewed all changes.

I am marking "Request for changes" just not to merge this by mistake without a new review.

In case somebody from Coreint reviews this and marks it as approved, don't hesitate to ignore my review :)

gsanchezgavier:
@kang-makes  you can just remove approval not to block the merging if you not consider that there are something to change

gsanchezgavier:
> Too many change since my last review
@kang-makes  Sorry for the force push , but main changes after you review are [this](https://github.com/newrelic/release-toolkit/pull/152/files#diff-3435321f89580c19c1ccf925d4992905d93df0e2f863bff8b1ad6261e0cdb7d2R65-R71) since it was not working at that moment

contrib/ohi-release-notes/action.yml
```diff
@@ -10,9 +10,6 @@ inputs:
   fail-if-held:
     description: Fail if the held toggle is active. Defaults to `true`.
     default: 'true'
-  link-dependencies-dictionary:
-    description: Sets the link dependency dictionary.
-    default: ''
```
kang-makes, line 15:
```suggestion
  link-dependencies-dictionary:
    description: Sets the link dependency dictionary.
    default: '.github/rt-dictionary.yml'
```
Even if it is an OHI, I would let users modify it.

@alvarocabanas posted this in an internal blog post and we expect people to start using this, and they might use our contrib action that is ready to be used.

contrib/ohi-release-notes/action.yml
```diff
@@ -65,11 +62,19 @@ runs:
...
+      shell: bash
+      run: |
+        cp ${{ github.action_path }}/rt-dictionary.yml /home/runner/work/_temp/_github_home/rt-dictionary.yml
     - name: Link dependencies
       if: ${{ steps.check-release.outputs.skip != 'true' }}
       uses: newrelic/release-toolkit/link-dependencies@v1
       with:
-        dictionary: ${{ inputs.link-dependencies-dictionary }}
+        # see comment from last step regarding this file.
+        dictionary: /github/home/rt-dictionary.yml
```
kang-makes, line 77:
```suggestion
        dictionary: ${{ inputs.link-dependencies-dictionary }}
```
Because of the comment I left above.

contrib/ohi-release-notes/action.yml
```diff
@@ -65,11 +62,19 @@ runs:
       shell: bash
       run: |
         echo "::set-output name=skip::${{ steps.empty.outputs.is-empty == 'true' || steps.held.outputs.is-held == 'true' }}"
+    # rt-dictionary is a file contained by this action and is located inside $GITHUB_ACTION_PATH, but this
+    # folder is not mounted in the container that runs the link-dependencies action, so we copied the file to
+    # /home/runner/work/_temp/_github_home/ which is mounted in /github/home/
+    - name: Copy dictionary to temp folder mounted by the link-dependencies action
+      shell: bash
+      run: |
+        cp ${{ github.action_path }}/rt-dictionary.yml /home/runner/work/_temp/_github_home/rt-dictionary.yml
```
kang-makes, line 71: I don't like to have these kinds of absolute paths that might not be standard because they can be changed by Github without notice as they seem to be internal.

Take a look at this issue from a year ago: https://github.com/actions/checkout/issues/785#issuecomment-1133397851
Runners were run in `/opt/actions-runner/_work/`

If the image that is built for this step only has access to `${GITHUB_WORKSPACE}`, let's use it.

```suggestion
        if ! [ -f "${GITHUB_WORKSPACE}/${{ inputs.link-dependencies-dictionary }}" ]; then
                cp ${{ github.action_path }}/rt-dictionary.yml "${GITHUB_WORKSPACE}/${{ inputs.link-dependencies-dictionary }}"
        fi
```

This way, we have a default for users to override our dictionary and we set a default in case it does not exist for us.

contrib/ohi-release-notes/run.sh
```diff
@@ -65,7 +68,7 @@ EOM
 # parsing flags
 EXCLUDED_DIRECTORIES=".github"
 IS_HELD_FAIL="--fail"
-DICTIONARY=".github/rt-dictionary.yaml"
+DICTIONARY=""
```
kang-makes, line 71: If the new default is going to be empty, the test below ([line 118, sorry, I cannot leave this comment there](https://github.com/newrelic/release-toolkit/pull/152/files#diff-611cd2df9226104e98791e3c12c2f96eb8e741fbf973033eb4b7d4c6abf8e019R118-R122)) instead of testing if the file exists should test that the string is empty.

If it is not empty, it should use the flag `--dictionary` and so `rt` fail explicitly if the file does not exist.

contrib/ohi-release-notes/rt-dictionary.yml
```diff
@@ -0,0 +1,31 @@
...
+  newrelic/nri-nginx: "https://github.com/newrelic/nri-nginx/releases/tag/{{.To}}"
+  newrelic/nri-oracledb: "https://github.com/newrelic/nri-oracledb/releases/tag/{{.To}}"
+  newrelic/nri-postgresql: "https://github.com/newrelic/nri-postgresql/releases/tag/{{.To}}"
+  newrelic/nri-prometheus: "https://github.com/newrelic/nri-prometheus/releases/tag/{{.To}}"
+  newrelic/nri-rabbitmq: "https://github.com/newrelic/nri-rabbitmq/releases/tag/{{.To}}"
+  newrelic/nri-redis: "https://github.com/newrelic/nri-redis/releases/tag/{{.To}}"
+  newrelic/nri-varnish: "https://github.com/newrelic/nri-varnish/releases/tag/{{.To}}"
+  newrelic/nri-winservices: "https://github.com/newrelic/nri-winservices/releases/tag/{{.To}}"
+  newrelic/nrjmx: "https://github.com/newrelic/nrjmx/releases/tag/{{.To}}"
+  newrelic/prometheus-exporter-supervisor: "https://github.com/newrelic/prometheus-exporter-supervisor/releases/tag/{{.To}}"
```
kang-makes, line 31: NIT: I have just realised that we do not have Prometheus exporters packages: https://github.com/newrelic/newrelic-prometheus-exporters-packages

Does it even make sense to have it?

contrib/ohi-release-notes/run.sh
```diff
@@ -3,13 +3,16 @@
 set -euo pipefail
 
 RT_PKG="github.com/newrelic/release-toolkit@latest"
+# TODO this is just added to check the CI tests are passing and should be reverted before merge , chicken egg thing.
+# DICTIONARY_URL="https://raw.githubusercontent.com/newrelic/release-toolkit/v1/contrib/ohi-release-notes/rt-dictionary.yml"
+DICTIONARY_URL="https://raw.githubusercontent.com/newrelic/newrelic-infra-checkers/main/rt-dictionary/.rt-dictionary.yml"
```
kang-makes, line 8: Smart move here :)

kang-makes (commented):
> Sorry for the force push.

No worries, this was a tricky PR that is getting out of hand.

I took this opportunity to review it deeper after seeing the issues you had and left a new set of comments.

All of them are open to discussion, let me know what you think :)

contrib/ohi-release-notes/action.yml
gsanchezgavier: I think that the scope of this action is for ohi, that would involve an extra complexity that I'm not sure that is needed at this point, it could be added in the future I think ( if needed) ,We could probably all benefit from the same dictionary at the end.

contrib/ohi-release-notes/action.yml
gsanchezgavier: sounds good, I will do a quick check to see if that works

contrib/ohi-release-notes/rt-dictionary.yml
gsanchezgavier: i don't think so, it shouldn't be a dependency anywhere. We could add it in the future if needed

contrib/ohi-release-notes/run.sh
gsanchezgavier: that test is done before, and if empty it gets the default dictionary, so it could never be an empty string.

contrib/ohi-release-notes/action.yml
gsanchezgavier: so finally done in [18c1d36](https://github.com/newrelic/release-toolkit/pull/152/commits/18c1d36223c4664ded6407f4c027b892c1c4d00b), it was actually less complex than I imagine, thanks for the sugestion.

contrib/ohi-release-notes/action.yml
gsanchezgavier: thanks! [18c1d36](https://github.com/newrelic/release-toolkit/pull/152/commits/18c1d36223c4664ded6407f4c027b892c1c4d00b)

contrib/ohi-release-notes/action.yml
```diff
@@ -11,8 +11,8 @@ inputs:
     description: Fail if the held toggle is active. Defaults to `true`.
     default: 'true'
   link-dependencies-dictionary:
-    description: Sets the link dependency dictionary.
-    default: ''
+    description: Sets the link dependency dictionary. By default common a dictionary is used.
```
sigilioso, line 14: IMO the `link-dependencies-dictionary` value can be a bit misleading.

We state that a common dictionary is used by default, but that is only true if `.github/rt-dictionary.yml` (the default value) does not exist in the repository, right? In other words, if no value is set for `link-dependencies-dictionary` but the `.github/rt-dictionary.yml` file does exist it will be used instead of the default being stored in the action.

I don't think there is a simpler way to keep it configurable at the same time we use the dictionary stored in the action by default so, maybe updating the field description is enough. WDYT?

```suggestion
    description: Sets the link dependency dictionary. `.github/rt-dictionary.yml` is used by default, if it does not exist a common dictionary is used.
```

contrib/ohi-release-notes/run.sh
sigilioso: We may even revert it **after** merging (in a different PR) in order to avoid by-passing merge restrictions

sigilioso (approved):
LGTM!

I simply left a suggestion regarding the description of one of the action attributes.

contrib/ohi-release-notes/run.sh
kang-makes: In that case, `$DICTIONARY` is not going to be empty. So the test in line [118](https://github.com/newrelic/release-toolkit/pull/152/files#diff-611cd2df9226104e98791e3c12c2f96eb8e741fbf973033eb4b7d4c6abf8e019R118-R122) does not make sense anymore, right?

kang-makes (approved):
The things that are missing are nits.

LGTM. Thank you very much! It looks awesome :)

contrib/ohi-release-notes/action.yml
gsanchezgavier: yea that comment is more clear thanks! I'm not worried about this since I don't think is going to be used tbh, I would prefer not to add any complexity because of it for maintainability.

contrib/ohi-release-notes/run.sh
gsanchezgavier: it has the same sense as it was before, to check if the specified file exists before passing it to rt

gsanchezgavier: oh now it make sesnse againt to have that default so is aligned with the action. i will change it

gsanchezgavier: done the defaults now match , between action and script

PR #156, opened 2023-05-23
Include exclude files

paologallinaharbur:
Supersedes https://github.com/newrelic/release-toolkit/pull/155 due to https://github.com/golang/go/issues/34175 (branchName cannot have `/` and therefore I could not test changes before merging)

Fixes also the bug shown in https://github.com/newrelic/release-toolkit/pull/157
When running in docker having the argument "--included-dirs=" causes to get an "empty" includeDir definition
`time="2023-05-24T15:25:42Z" level=error msg="includedDirs = [\"\"]"`


contrib/ohi-release-notes/action.yml
```diff
@@ -7,6 +7,9 @@ inputs:
   excluded-dirs:
     description: Exclude commits whose changes only impact files in specified dirs relative to repository root. Defaults to ".github".
     default: '.github'
+  excluded-files:
+    description: Exclude commits whose changes only impact files in specified files relative to repository root. Defaults to "".
+    default: ''
```
sigilioso, line 12: Would it make sense to expose the included files and/or dirs in the `ohi-release-notes` action as well?

src/app/generate/generate.go
```diff
@@ -59,10 +61,22 @@ var Cmd = &cli.Command{
 		},
 		&cli.StringSliceFlag{
 			Name:    excludedDirsFlag,
-			EnvVars: common.EnvFor(includedDirsFlag),
+			EnvVars: common.EnvFor(excludedDirsFlag),
```
sigilioso, line 64: Thanks for noticing and addressing this!

src/git/commit_filter.go
```diff
@@ -79,7 +119,7 @@ func (s *CommitFilter) Commits(lastHash string) ([]Commit, error) {
 
 // commitChangesIncluded returns true if at least one of the changes is included in includedDirs.
 func (s *CommitFilter) commitChangesIncluded(files []string) bool {
-	if len(s.includedDirs) < 1 {
+	if len(s.includedDirs) < 1 && len(s.includedFiles) < 1 {
```
sigilioso, line 122: Nit: we may want to update the function commit to include `includedFiles` or just make the comment more generic.

src/git/commit_filter.go
```diff
@@ -89,14 +129,20 @@ func (s *CommitFilter) commitChangesIncluded(files []string) bool {
...
+		}
 	}
 
 	return false
 }
 
 // commitChangesExcluded returns true if all changes are in excludedDirs.
 func (s *CommitFilter) commitChangesExcluded(files []string) bool {
-	if len(s.excludedDirs) < 1 {
+	if len(s.excludedDirs) < 1 && len(s.excludedFiles) < 1 {
```
sigilioso, line 145: Nit: the same that the previous comment.

src/git/commit_filter_test.go
```diff
@@ -185,10 +185,20 @@ func TestCommitFilter_NewCommitFilter(t *testing.T) {
 			},
 			expectedError: git.ErrDirNotValid,
 		},
+		{
+			name: "Directory_for_files_Throw_No_Error",
```
sigilioso, line 189: 🤔 but it is throwing an error, isn't it?

sigilioso (commented):
Thanks for taking care of this! It looks really good.
I left some comments and questions (merely nits)

src/git/commit_filter_test.go
paologallinaharbur: copy/paste 😮‍💨

contrib/ohi-release-notes/action.yml
paologallinaharbur: I would keep the list as small as possible. Since we do not need them I would not add those for now.

I added them to the command just for "simmetry"

contrib/ohi-release-notes/run.sh
```diff
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-RT_PKG="github.com/newrelic/release-toolkit@latest"
+# TODO
```
paologallinaharbur, line 5: TODO, just not to forget about this

contrib/ohi-release-notes/action.yml
sigilioso: Maybe I didn't understand it well, but I thought we already had a use case for `include-files`. Don't we want to take into account only changes in [infrastructure-bundle/bundle.yml](https://github.com/newrelic/infrastructure-bundle/blob/master/bundle.yml)?

contrib/ohi-release-notes/run.sh
```diff
@@ -65,6 +67,7 @@ EOM
 
 # parsing flags
 EXCLUDED_DIRECTORIES=".github"
+EXCLUDED_FILES=""
```
kang-makes, line 70:
```suggestion
EXCLUDED_DIRECTORIES_FLAG="--excluded-dirs=.github"
EXCLUDED_FILES_FLAG=""
```

contrib/ohi-release-notes/run.sh
```diff
@@ -80,6 +83,7 @@ while true; do
             --git-root ) GIT_ROOT="$2"; shift 2 ;;
             # Flags for generate-yaml
             --excluded-dirs ) EXCLUDED_DIRECTORIES="$2"; shift 2 ;;
+            --excluded-files ) EXCLUDED_FILES="$2"; shift 2 ;;
```
kang-makes, line 86:
```suggestion
            --excluded-dirs ) EXCLUDED_DIRECTORIES_FLAG="--excluded-dirs=$2"; shift 2 ;;
            --excluded-files ) EXCLUDED_FILES_FLAG="--excluded-files=$2"; shift 2 ;;
```

contrib/ohi-release-notes/run.sh
```diff
@@ -110,7 +114,7 @@ fi
 
 
     # generating the changelog
-    ${RT_BIN} generate-yaml --excluded-dirs "$EXCLUDED_DIRECTORIES"
+    ${RT_BIN} generate-yaml --excluded-dirs "$EXCLUDED_DIRECTORIES" --excluded-files "$EXCLUDED_FILES"
```
kang-makes, line 117: With the change that we are, we must have the use case of the excluded files and/or directories flags being empty.

```suggestion
    ${RT_BIN} generate-yaml "$EXCLUDED_DIRECTORIES_FLAG" "$EXCLUDED_FILES_FLAG"
```

kang-makes (commented):
Thanks for taking care of this. It is looking good!

There are no [self-tests for the action](https://github.com/newrelic/release-toolkit/blob/includeExcludeFiles/.github/workflows/self_test.yaml#L124-L177) that use the new flags, right?

contrib/ohi-release-notes/action.yml
paologallinaharbur: we said we were going to exclude only `go.sum` and `go.mod`, @kang-makes since you add a 👍 would you like as well for the infra-bundle to have just the include instead?

contrib/ohi-release-notes/run.sh
paologallinaharbur: I do not thinks it is an issue:
```
rt generate-yaml --included-dirs --excluded-dirs
INFO[0000] Ignoring 0.0.2 since it belongs to a different branch 
INFO[0000] Renovate source did not find any commit since "190097158a3dda47531b636f81392f587d233faf" 
INFO[0001] Ignoring 0.0.2 since it belongs to a different branch 
INFO[0001] Dependabot source did not find any commit since "190097158a3dda47531b636f81392f587d233faf" 
changelog is empty

```

contrib/ohi-release-notes/action.yml
paologallinaharbur: I am reluctant to add not strictly needed options to the contrib action since there is the risk of having it growing and growing. Since it is like an action grouping all the other ones, we could easily end-up with "tens" of them :)

contrib/ohi-release-notes/run.sh
paologallinaharbur: and it is in two places

contrib/ohi-release-notes/action.yml
kang-makes: I know that the contrib action might finish with lots of parameters and options but it should cover our use cases.

We have use-cases where we want to exclude the whole folder (like `.github` or `tools`) but we also have repositories where we want to exclude everything but a file (like the `bundle.yaml` of the infra bundle or `exporter.yaml` files in the exporter packages repository).

As we have many use cases, I would cover both: `exclude-directories` and `include-files` at least.

contrib/ohi-release-notes/run.sh
kang-makes: Hhmm... I have a question then. Shouldn't then be mutual exclusive flags?

If we set `include` we should only have into account commits that affect that file. What is the use case to have both?

paologallinaharbur: Not sure if I understood the question:
You could have something like exclude dir "folder1" and exclude file "folder2/file"

contrib/ohi-release-notes/action.yml
paologallinaharbur: By the way, since you are both interested into it and it is not that much toil I'll add them.


..... But I still think they will be unused 🤣

src/app/generate/generate.go
```diff
@@ -216,3 +236,18 @@ func tagVersionGetter(cCtx *cli.Context) (*git.TagsSource, error) {
 
 	return tvsrc, nil
 }
+
+func sanitizeValue(in []string) []string {
+	// The action leverages docker, but having no default the argument "--included-dirs=" causes
+	// to get an "empty" includeDir definition, but with "" as an element
+	// time="2023-05-24T15:25:42Z" level=error msg="includedDirs = [\"\"]"
+	// This causes "" to be passed as an element, that is translated to . and a slash is added becoming "./".
+	// We have an issue there since "./" matches all files on root, but not any of the folders that are specified as "folder1/..".
```
sigilioso, line 245: nit: the comment is great for context but maybe is a bit too coupled with the way the issue was reproduced.
I wonder if we could include a test reproducing the issue and move the long explanation there. Here we could simply point out that `""` would become `"./"` with unexpected behavior.

src/app/generate/generate.go
```diff
@@ -216,3 +236,18 @@ func tagVersionGetter(cCtx *cli.Context) (*git.TagsSource, error) {
...
+
+	if len(in) == 1 {
+		if in[0] == "" {
+			return []string{}
+		}
+	}
+	return in
```
sigilioso, line 252: Would values such as `["", "foo/bar"]` (having more than one element but containing empty strings) be valid?

paologallinaharbur: I'll add it to the validation

src/app/generate/generate_test.go
```diff
@@ -331,6 +331,41 @@ dependencies:
       meta:
         commit: chore(deps): bump thisdep from 1.7.0 to 1.10.1`) + "\n",
 		},
+		{
+			name:   "Markdown_Dependabot_Filter_IncludedFiles",
```
sigilioso, line 335: nit: I'd update the name to make it different from the previous one.

src/git/commit_filter.go
```diff
@@ -77,7 +77,7 @@ func ExcludedFiles(excludedFiles ...string) CommitFilterOptionFunc {
 }
 
 func isDirNameInvalid(dir string) bool {
-	return dir == "." || dir == ".." || strings.HasPrefix(dir, "/") || strings.Contains(dir, "./")
+	return dir == "." || dir == ".." || dir == "" || strings.HasPrefix(dir, "/") || strings.Contains(dir, "./")
```
sigilioso, line 80: Maybe it is too much but (because I fear we are relying on integration tests only), would it make sense to include an additional test to check the new validation? It would have `--included-files=invented,` or something similar.

sigilioso (approved):
LGTM 👍 
I left a couple of _nits_ regarding tests.

src/git/commit_filter.go
paologallinaharbur: I can add one having `["invented", ""]` and seeing that is refused,

contrib/ohi-release-notes/run.sh
kang-makes: Take this as an example:
```
.github
go.mod
bundle.yaml
```

If I include `bundle.yaml` it makes no sense to "exclude `.github`" because we are excluding everything except `bundle.yaml`, right?

That is why I am asking if these are mutually exclusive. We should not allow using `exclude` and `include` flags at the same time.

paologallinaharbur: Since one is an allowList and the other is a denyList I think that the use-case to have them at the same time if whenever you have conflicts:
```
/allow-this
/allow-this/without-this
```

In the case above you can allow everything in `/allow-this` but not in `/allow-this/without-this`.

paologallinaharbur: When allowing/not-allowing files only I think you are right, it is not useful to have both options at the same time.

kang-makes: The problem then is even worse. Which is the order?
What if I want to exclude a directory and include s ingle file of that folder?

From my point of view, it is better to just disallow to use both of the flags at the same time and take time to implement a good ignore mechanism like `.gitignore`, `.helmignore`, or `.dockerignore`.

paologallinaharbur: > take time to implement a good ignore mechanism
I fear that doing it would be a total overkill given our needs (Honestly, I would not have the include option at all and keep it simple)

Regarding the "order", it is specified in the comments: since we are defining filters, `exclude` takes precedence right 
now. If it is included and excluded at the same time -> it is excluded

Basically, a file is included if:
 - (it is not excluded) &&  (if any include rule is specified - it is included)

paologallinaharbur: I can document it in the `generate-yaml` readme if you think that having both is not that un-common

src/git/commit_filter_test.go
```diff
@@ -41,74 +41,143 @@ func TestCommitFilter_Commits(t *testing.T) {
...
 			opts: []git.CommitFilterOptionFunc{
 				git.IncludedDirs("folder1"),
 			},
-			commits: allCommits,
 			expectedCommits: []git.Commit{
 				singleFileFolder1, threeFilesFolders, rootAndFolder,
 			},
 		},
+		{
+			name: "Include_folder1/single-file-on-folder1",
```
gsanchezgavier, line 62: nit: perhaps a more descriptive name like `Include commits that only modifies IncludedFiles`

gsanchezgavier (approved):
LGTM I left some nits. Thanks

README_CLI.md
```diff
@@ -14,16 +14,18 @@ Builds a changelog.yaml file from multiple sources.
...
+| `included-dirs`  |                | Only scan commits scoping at least one file in any of the following comma-separated directories, relative to repository root (--dir) (Paths may not start with "/" or contain ".." or "." tokens)                         |
+| `excluded-dirs`  |                | Exclude commits whose changes only impact files in specified dirs relative to repository root (--dir) (separated by comma) (Paths may not start with "/" or contain ".." or "." tokens)                                   |
+| `included-files` |                | Only scan commits scoping at least one file in any of the following comma-separated ones, relative to repository root (--dir) (Paths may not start or end with "/" or contain ".." or "." tokens)                         |
+| `excluded-files` |                | Exclude commits whose changes only impact the specified files, path are relative to repository root (--dir) (separated by comma) (separated by comma) (Paths may not start or end with "/" or contain ".." or "." tokens) |
```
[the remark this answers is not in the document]
paologallinaharbur: ok I added it in the general readme

contrib/ohi-release-notes/run.sh
paologallinaharbur: documented!

PR #159, opened 2023-06-06
feat(render): dedup dependencies

paologallinaharbur:
This PR aims to reduce noise and simplify the changelog to avoid situations like the following.

Screenshot 2023-06-06 at 11 39 29

The current solution is not keeping the "From" value from the very first dep bump encountered, if we think it is useful we can add a simple logic to keep it.

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,10 @@ The list of entry types, and therefore groups, supported by the release-toolkit 
 
 Additionally, a section with changes to dependencies is also included after the list of changes.
 
+Notice that the list of dependencies is deduplicated in case the very same dependency is updated more than once.
+Merely the latest bump per each dependency is shown in order to simplify the changelog.
+
+
 The Markdown version of the `changelog.yaml` file mentioned above would be:
 
 ```markdown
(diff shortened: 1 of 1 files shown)


paologallinaharbur:
blocked by https://github.com/newrelic/newrelic-infra-checkers/compare/fix/disableDepguard?expand=1

src/app/render/render.go
```diff
@@ -72,7 +72,7 @@ func Render(cCtx *cli.Context) error {
 		return fmt.Errorf("creating destination file at %q: %w", mdPath, err)
 	}
 
-	rnd := renderer.New(ch)
+	rnd := renderer.New(ch, renderer.WithDeduplicateDependencies())
```
sigilioso, line 75: Since it is always enabled, should we avoid including the `renderer.WithDeduplicateDependencies()` option and make it the default behavior?

src/changelog/renderer/renderer.go
```diff
@@ -85,9 +99,32 @@ func (r Renderer) parse() parsedChangelog {
...
+	}
+
+	for _, dep := range deps {
 		parsed.Sections[string(changelog.TypeDependency)] = append(parsed.Sections[string(changelog.TypeDependency)], dep)
 	}
 
 	return parsed
 }
+
+// Dependencies are ordered. We keep the latest that should be the one with the latest semVer.
```
sigilioso, line 114: nit:
```suggestion
// Dependencies are sorted in ascending order. We keep the latest that should be the one with the latest semVer.
```

sigilioso (commented):
I like the approach 👍

I've simply left a comment and a _nit_.

I would also include a note explaining that dependency changes are _deduplicated_, an small example like the one in the new test would be almost self-explanatory.

sigilioso (approved):
LGTM 🚀 

Thanks for taking care of it and addressing the comments.

It may be helpful to include a comment in the [README's render section](https://github.com/newrelic/release-toolkit/blob/main/README.md#render-markdown-and-update-markdown) to explain this behavior.

PR #160, opened 2023-06-07
feat(generate): exclude dev dependencies (NR-122276)

marcsanmi:
This PR adds the capability of excluding specific development libraries from the CHANGELOG.md. 

It adds a default exclude-dev-dependencies file to be used by the action so for now, there's no need to update the trigger prerelease workflow.


src/git/commit_filter.go
```diff
@@ -111,7 +119,7 @@ func (s *CommitFilter) Commits(lastHash string) ([]Commit, error) {
 
 	filteredCommits := make([]Commit, 0)
 	for _, commit := range commits {
-		if !s.commitChangesExcluded(commit.Files) {
+		if !s.commitChangesExcluded(commit.Files) && !s.commitDevDependenciesExcluded(commit.Message) {
```
sigilioso, line 122: nit: I wonder if, now that we have more than one filter, we should rename `commitChangesExcluded` to `commitExcludedByFiles` (just to make each function responsibility more explicit)

sigilioso (commented):
Thanks! The changes look really good!

I've left some comments and suggestions.

There are also some missing EOF in test/sample files.

On the other hand, should we include this new configuration option in the README as well?

contrib/ohi-release-notes/excluded-dev-dependencies.yml
```diff
@@ -0,0 +1,5 @@
+# This contains a list of development libraries that should be skipped from the changelog.
+excluded_dev_dependencies:
+  - github.com/stretchr/testify
+  - github.com/testcontainers/testcontainers-go
+  - "feat(generate): include/exclude files"
```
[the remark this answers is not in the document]
marcsanmi: This should be removed 😓

src/git/commit_filter.go
marcsanmi: Yes since the `commitChangesExcluded ` it's generic to excluded changed; thanks for the suggestion 👍

marcsanmi: Done in 1c488f32c1c81e2107b13ffb8801c864bbefaa8e.

sigilioso (approved):
LGTM

src/app/common/devdependencies.go
kang-makes, line 1: File is called `devdependencies` and the struct inside `ExcludedDependencies`. Can we change it so they match?

src/app/generate/generate.go
```diff
@@ -182,7 +197,7 @@ func Generate(cCtx *cli.Context) error {
 	return nil
 }
 
-func addDepSource(cCtx *cli.Context, sources []changelog.Source, appendDep appendDepSrc) ([]changelog.Source, error) {
+func addDepSource(cCtx *cli.Context, sources []changelog.Source, appendDep appendDepSrc, deps []string) ([]changelog.Source, error) {
```
kang-makes, line 200:
```suggestion
func addDepSource(cCtx *cli.Context, sources []changelog.Source, appendDep appendDepSrc, excludedDeps []string) ([]changelog.Source, error) {
```

kang-makes (commented):
I would not do the contrib action change in this PR if we are not able to test it properly in the same PR.

On the other hand, looks really good :)

src/app/common/devdependencies.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+// ExcludedDependencies represents the structure of the YAML file.
+type ExcludedDependencies struct {
```
paologallinaharbur, line 11: NIT: is it exported due to yaml Unmarshal? Aren't just the fields forced to be exported?

src/app/generate/generate.go
```diff
@@ -240,10 +257,9 @@ func tagVersionGetter(cCtx *cli.Context) (*git.TagsSource, error) {
 func sanitizeValue(in []string) []string {
 	// Even if the user passes "an empty string it is considered as an element "",
 	// and translated to "./" causing an unexpected behavior.
-	if len(in) == 1 {
-		if in[0] == "" {
-			return []string{}
-		}
+	if len(in) == 1 && in[0] == "" {
```
paologallinaharbur, line 260: if len(in)==0, in[0] could cause issues?

src/app/generate/generate_test.go
```diff
@@ -396,6 +396,64 @@ dependencies:
...
+		{
+			name:   "Markdown_Dependabot_Filter_ExcludedDependencies",
+			md:     mdChangelog,
+			args:   fmt.Sprintf("--renovate=false --excluded-dependencies=%s", path.Join("..", "testdata", "excluded-dependencies.yml")),
+			author: "dependabot <dependabot@github.com>",
+			commits: []string{
+				"chore(deps): bump github.com/stretchr/testify from 1.7.0 to 1.8.0",
+				"chore(deps): bump thisdep from 1.7.0 to 1.10.1",
+			},
+			// Note: meta.commit is actually the commit hash.
```
paologallinaharbur, line 437: does it make sense to copy this message each time?

marcsanmi: Probably not, since it was in the previous 10 subtests I didn't want to "break with the standard" 😁 I'm going to remove it from all of them except the 1st.

paologallinaharbur:
> I would not do the contrib action change in this PR if we are not able to test it properly in the same PR.
Why can't we check it pointing to the local branch? I did it [here](https://github.com/newrelic/release-toolkit/pull/156/commits/0847036e3a0289bf8e064ab6de8b250c884deae7) for both the action and the script (you cannot have `/` in the tagname)

contrib/ohi-release-notes/action.yml
```diff
@@ -53,13 +56,23 @@ runs:
...
         markdown: ${{ inputs.git-root }}/CHANGELOG.md
+    # excluded-dependencies is a file contained by this action and is located inside $GITHUB_ACTION_PATH, but this
+    # folder is not mounted in the container that runs the generate-yaml action, so we copied the file to
+    # ${GITHUB_WORKSPACE} which is mounted in the container.
+    - name: Copy dependencies file to temp folder mounted by the generate action
+      shell: bash
+      run: |
+        if ! [ -f "${GITHUB_WORKSPACE}/${{ inputs.excluded-dependencies }}" ]; then
+          cp ${{ github.action_path }}/excluded-dependencies.yml "${GITHUB_WORKSPACE}/${{ inputs.excluded-dependencies }}"
+        fi
```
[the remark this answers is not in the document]
marcsanmi: As discussed offline, for now, we'll go with just this PR, but for future contributions, it's better to re-tag first `v1` and then update the ohi contrib action in a follow-up PR. In this case it was tested but pointing to this branch as commented https://github.com/newrelic/release-toolkit/pull/160#issuecomment-1580794953, but the PR will be merged failing since will point to v1.

paologallinaharbur (approved):

src/app/generate/generate.go
```diff
@@ -17,16 +17,17 @@ import (
...
+	markdownPathFlag         = "markdown"
+	renovateFlag             = "renovate"
+	dependabotFlag           = "dependabot"
+	tagPrefixFlag            = "tag-prefix"
+	gitRootFlag              = "git-root"
+	includedDirsFlag         = "included-dirs"
+	excludedDirsFlag         = "excluded-dirs"
+	includedFilesFlag        = "included-files"
+	excludedFilesFlag        = "excluded-files"
+	excludedDependenciesFlag = "excluded-dependencies"
```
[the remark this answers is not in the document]
paologallinaharbur: Also "excluded-dependencies-path" could work

contrib/ohi-release-notes/action.yml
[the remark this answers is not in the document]
paologallinaharbur: last time I tested it changing the tag pointing to the local branch and then changing it back

marcsanmi:
> Why can't we check it pointing to the local branch? I did it [here](https://github.com/newrelic/release-toolkit/pull/156/commits/0847036e3a0289bf8e064ab6de8b250c884deae7) for both the action and the script (you cannot have / in the tagname)

Yes, It's what I did here https://github.com/newrelic/release-toolkit/pull/160#issuecomment-1580794953 but then we need to change it back to `v1`.  So, we don't have to merge with a not passing pipeline

.github/workflows/self_test.yaml
```diff
@@ -608,7 +613,7 @@ jobs:
           git-root: ${{ env.MOCK_REPO_ACTION }}
           excluded-files: dependencies-skip1
           excluded-dirs: dependencies-repo-skip
-          included-dirs: dependencies-repo
+          included-dirs: dependencies-repo,excluded-dependencies-repo
```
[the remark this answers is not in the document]
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

src/app/common/devdependencies.go
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

src/app/generate/generate.go
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

src/app/generate/generate.go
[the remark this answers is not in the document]
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

contrib/ohi-release-notes/run.sh
```diff
@@ -119,7 +129,11 @@ fi
 
 
     # generating the changelog
-    ${RT_BIN} generate-yaml "$EXCLUDED_DIRECTORIES_FLAG" "$EXCLUDED_FILES_FLAG" "$INCLUDED_DIRECTORIES_FLAG" "$INCLUDED_FILES_FLAG"
+    if [ -f "$EXCLUDED_DEPENDENCIES" ]; then
+      ${RT_BIN} generate-yaml "$EXCLUDED_DIRECTORIES_FLAG" "$EXCLUDED_FILES_FLAG" "$INCLUDED_DIRECTORIES_FLAG" "$INCLUDED_FILES_FLAG" "$EXCLUDED_DEPENDENCIES"
+    else
```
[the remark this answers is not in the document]
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

src/app/generate/generate_test.go
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

src/app/generate/generate.go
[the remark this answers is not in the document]
marcsanmi: Done in 4ffdd63741f6aca325bc28144f13abf165546015.

marcsanmi:
The latest pipeline passing can be found here https://github.com/newrelic/release-toolkit/actions/runs/5208478495/jobs/9397116030?pr=160. Switching back to `v1`...

contrib/ohi-release-notes/action.yml
```diff
@@ -53,13 +56,23 @@ runs:
...
+    # ${GITHUB_WORKSPACE} which is mounted in the container.
+    - name: Copy dependencies file to temp folder mounted by the generate-yaml action
+      shell: bash
+      run: |
+        if ! [ -f "${GITHUB_WORKSPACE}/${{ inputs.excluded-dependencies-manifest }}" ]; then
+          cp ${{ github.action_path }}/excluded-dependencies.yml "${GITHUB_WORKSPACE}/${{ inputs.excluded-dependencies-manifest }}"
+        fi
     - name: Generate YAML
-      uses: newrelic/release-toolkit/generate-yaml@v1
+      uses: newrelic/release-toolkit/generate-yaml@msanmiquel/exclude-dev-dependencies
```
kang-makes, line 69: This is still pointing to the branch.

contrib/ohi-release-notes/run.sh
```diff
@@ -4,6 +4,7 @@ set -euo pipefail
 
 RT_PKG="github.com/newrelic/release-toolkit@latest"
 DICTIONARY_URL="https://raw.githubusercontent.com/newrelic/release-toolkit/v1/contrib/ohi-release-notes/rt-dictionary.yml"
+EXCLUDED_DEPENDENCIES_MANIFEST_URL="https://raw.githubusercontent.com/newrelic/release-toolkit/msanmiquel/exclude-dev-dependencies/contrib/ohi-release-notes/excluded-dependencies.yml"
```
kang-makes, line 7: ditto

contrib/ohi-release-notes/run.sh
```diff
@@ -119,7 +129,7 @@ fi
 
 
     # generating the changelog
-    ${RT_BIN} generate-yaml "$EXCLUDED_DIRECTORIES_FLAG" "$EXCLUDED_FILES_FLAG" "$INCLUDED_DIRECTORIES_FLAG" "$INCLUDED_FILES_FLAG"
+    ${RT_BIN} generate-yaml "$EXCLUDED_DIRECTORIES_FLAG" "$EXCLUDED_FILES_FLAG" "$INCLUDED_DIRECTORIES_FLAG" "$INCLUDED_FILES_FLAG" "$EXCLUDED_DEPENDENCIES_MANIFEST"
```
kang-makes, line 132: This is not working as expected. That is the reason I gave @paologallinaharbur the tip of calling variables `$WHATEVER_FLAG`.

~~you are using the manifest file variable here, not the flag one. It is working because you are testing the defaults without passing a flag so at line 95 you are adding `--excluded-dependencies-manifest=` but in case a user sets its own file it stops working.~~ As 

This is hard to read, please, use a different variable for the file and the flag instead of reusing the same variable.

contrib/ohi-release-notes/run.sh
```diff
@@ -111,6 +115,12 @@ if ! [ -f "$DICTIONARY" ]; then
     curl -s -o "$DICTIONARY" "$DICTIONARY_URL"
 fi
 
+# fetch default excluded-dependencies-manifest by default
+if ! [ -f "$EXCLUDED_DEPENDENCIES_MANIFEST" ]; then
```
kang-makes, line 119: If a user uses `--excluded-dependencies-manifest`, at line 95 this variable is not going to be a file so the test `-f` is going to fail and you are going to download the file and overwrite the variable with a new file.

kang-makes (commented):
I am still not completely sure that the script works as expected.

We are testing the happy path and it works, but the usage of the flag is not being tested.

Self-test is not testing the flag being used and I think everything is going to break as soon as you try to use to exclude (for example) `nri-docker` from the dependencies.

src/app/generate/generate.go
```diff
@@ -123,11 +130,19 @@ func Generate(cCtx *cli.Context) error {
 	combinedChangelog := &changelog.Changelog{}
 	sources := make([]changelog.Source, 0)
 
+	var excludedDependencies []string
+	if excludedDependenciesPath := cCtx.String(excludedDependenciesFlag); excludedDependenciesPath != "" {
+		excludedDependencies, err = common.LoadExcludedDependencies(excludedDependenciesPath)
```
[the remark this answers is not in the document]
marcsanmi: You're right, if it's shared in the future we can place it in another package like `common`. Let's avoid "premature optimizations"... 👍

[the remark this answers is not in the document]
marcsanmi: Done in https://github.com/newrelic/release-toolkit/pull/160/commits/0c553d2c6d43b19e3d9fc26b7dccbedb2a8b26aa.

contrib/ohi-release-notes/action.yml
marcsanmi: I'm updating it at the end as I was "playing around" 😢   with the self test.

kang-makes (approved):
Thanks for the hard work. It looks awesome. Congrats :D

Issue #174, opened 2023-10-19
Inconsistent CHANGELOG link generation for dependency versions


marcsanmi:
The release toolkit isn't generating correct CHANGELOG links for dependencies in some cases. Examples of this can be seen in:

- https://github.com/newrelic/nri-ecs/releases/tag/v1.10.8
- https://github.com/newrelic/infrastructure-bundle/releases/tag/v3.2.22

The issue arises because the link is generated by appending the version number without the preceding `v`. This approach isn't consistent for the integrations, but it does work for the infrastructure agent where releases are created without the leading `v`. For instance:

- https://github.com/newrelic/infrastructure-bundle/releases/tag/v3.2.21

We need to think of a solution that works in both formats.

Related: https://github.com/newrelic/nri-ecs/issues/139

sigilioso:
It might be related to #134

paologallinaharbur:
The issue seems bigger than expected, different repos are affected: https://github.com/newrelic/infrastructure-bundle/releases/tag/v3.2.23

paologallinaharbur:
Fixed, the root cause was the dictionary

PR #175, opened 2023-10-31
fix(rt): leading v in dictionary

paologallinaharbur:
All releases changelog could be broken https://github.com/newrelic/release-toolkit/issues/174 whenever it was involving the dictionary (i.e. nri-ecs, infra bundle)


sigilioso (approved):
🚀 Thanks for taking care of this!

.github/workflows/self_test.yaml
```diff
@@ -584,7 +584,7 @@ jobs:
 
           ### ⛓️ Dependencies
           - Upgraded newrelic/infrastructure from 0.0.0 to 0.0.1 - [Changelog 🔗](https://github.com/newrelic/infrastructure-agent/releases/tag/0.0.1)
-          - Upgraded newrelic/nri-docker from 0.0.3 to 0.0.4 - [Changelog 🔗](https://github.com/newrelic/nri-docker/releases/tag/0.0.4)
+          - Upgraded newrelic/nri-docker from 0.0.3 to 1.8.26 - [Changelog 🔗](https://github.com/newrelic/nri-docker/releases/tag/v1.8.26)
```
sigilioso, line 587: Since it bumps a major now, I think the corresponding version above will also need a new major (v2.0.0 instead of v1.3.0)

paologallinaharbur: Moreover the test points to the local contrib action that points to V1, therefore it is a circular dependency with the release

PR #202, opened 2024-08-05
validator: check entry type through substring matching

### LINKED ISSUE #130: Validator is so strict with changelog
The validator is expecting a header with an exact match with the entry-type, while the generate-yaml isn't.
We should have a similar logic.

roobre:
Markdown validator tripped on me when using `Enhancements` (plural). As the `render` action outputs it in plural form, it only makes sense for the validator to also accept it in plural form.

Closes #130


kang-makes (approved):
Thanks for taking care of this :D

